PR #66, opened 2014-05-01
Autoload mu-plugins

Foxaii:
An autoloader that enables standard plugins to be required just like must-use plugins. The autoloaded plugins are included after all mu-plugins and standard plugins have been loaded. An asterisk (*) next to the name of the plugin designates the plugins that have been autoloaded.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ web/app/upgrade
 web/app/uploads/*
 !web/app/uploads/.gitkeep
 !web/app/mu-plugins/register-theme-directory.php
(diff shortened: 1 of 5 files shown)


olivierhaustrate:
I don't manage to get the autoloader to work on a subdomain with a WordPress single install. Any idea how to solve the problem?
On production server (Apache) I've set a subdomain (dev) for staging purpose. 
The plugin's files are loaded in the mu-plugins folder, but not there when I go to the plugin page in dashboard. 
If I try without the subdomain all is working great. So I understand that the problem comes from the subdomain install. 
Thanks for your help,
olivier

PR #110, opened 2014-10-29
Update bedrock-autoloader.php

chrisnx:
Call load_plugins() method on construct instead of the plugins_loaded hook

The mu-plugins are actioned before normal plugins in the action order. Some mu plugins are required
The bedrock-autoloader mu-plugin seems to fulfil two purposes. it a) allows the use of composer to make mu-plugins dependancies by running mu-plugins from within directories. b) allows normal plugins to act like mu plugins. However by running the mu-plugins further down the action order list, neither mu-plugins or plugins are behaving exactly as standard mu-plugins do as standard in wordpress. This could be important if the mu-plugin you are requiring via composer does some setup that normal plugins will then use. In the current implementation it will be loaded after the normal plugins.

### DIFF
--- a/web/app/mu-plugins/bedrock-autoloader.php
+++ b/web/app/mu-plugins/bedrock-autoloader.php
@@ -14,136 +14,136 @@ namespace Roots\Bedrock;
 if (!is_blog_installed()) { return; }
 
 class Autoloader {
-  private static $cache; // Stores our plugin cache and site option.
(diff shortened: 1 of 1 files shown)


christiansherry:
@swalkinshaw just realised the description for this plugin is out of date since this change, could that be updated too?

swalkinshaw:
@christiansherry good catch, just updated it.

PR #121, opened 2014-11-21
Set PHP version requirement back to >= 5.3.2. Fixes #120

### LINKED ISSUE #120: composer.json implies PHP >= 5.4 required, but README.md does not
Just wanted to check if PHP 5.4 is now a requirement of using bedrock? I've just been trying to grab version 1.3.0 and I'm unable to install it via composer. Taking a look in the README.md file I can see that you specify PHP >= 5.3.2 as a requirement (which I have, running 5.3.29) and I've not had trouble with previous versions.

Looks like the issue is coming from composer.json:

```
  "require": {
    "php": ">=5.4",
    "johnpbloch/wordpress": "4.0.1",
    "composer/installers": "v1.0.12",
    "vlucas/phpdotenv": "1.0.9"
  },
```

CharlieHawker:
This seems to solve the issue and still allow `composer update` / `composer install` to run without issue.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,7 +40,7 @@
     }
   ],
   "require": {
-    "php": ">=5.4",
+    "php": ">=5.3.2",
     "johnpbloch/wordpress": "4.0.1",
     "composer/installers": "v1.0.12",
     "vlucas/phpdotenv": "1.0.9"
--- a/composer.lock
+++ b/composer.lock
@@ -1,10 +1,9 @@
 {
     "_readme": [
         "This file locks the dependencies of your project to a known state",
-        "Read more about it at http://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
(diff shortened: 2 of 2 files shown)


DamonGant:
Id' just like to mention that distributions that were the most recent quite recently will be on 5.3 for a while. (e.g. wheezy for at least another 8 months, trusty for almost 2 years) There are a lot of sites on the shared host I must deploy on, so foreign packages or upgrading PHP is not an option so I'd encourage it if requirements aren't bumped up without a good reason.

swalkinshaw:
@DamonGant PHP 5.3 is end of life. I can't think of a better reason than that. In fact 5.4 is only getting security updates for another 3 months. So on September 14th we'll probably be bumping this to 5.5 as well.

retlehs:
https://php.net/eol.php

PR #133, opened 2015-01-09
use ~ operator on composer requirements

thewunder:
This commit uses the ~ operator on composer requirements so that composer upgrade grabs the latest bugfixes etc...

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,8 +41,8 @@
   ],
   "require": {
     "php": ">=5.4",
-    "composer/installers": "v1.0.12",
-    "vlucas/phpdotenv": "1.0.9",
+    "composer/installers": "~1.0.12",
+    "vlucas/phpdotenv": "~1.0.9",
     "johnpbloch/wordpress": "4.1.1"
   },
   "extra": {


mAAdhaTTah:
Worth noting [4.0.1 broke some plugins due to bad plugin dev practices](http://wptavern.com/wordpress-4-0-1-exposes-bad-development-practices-used-in-some-plugins). Not sure we want composer to auto upgrade that.

austinpray:
Definitely no `~` on wordpress. Wordpress does not follow semver anyway.

swalkinshaw:
@thewunder I'm fine with `~` for the non WP packages if you want to update this.

thewunder:
The composer.lock file will prevent it from being automatically upgraded without testing...

I'm mostly a Symfony dev not a wordpress dev, so if you don't think it's safe I'll remove.

jaywilliams:
This is the way I have mine setup, actually. Composer locks the version, so any version change can easily be reverted.

swalkinshaw:
@thewunder sorry kind of forgot about this. Would you be able to rebase it so it can be merged?

thewunder:
Rebased

retlehs:
@thewunder not actually rebased FYI (shouldn't be showing 13 commits on this PR, just 1)

thewunder:
That's what happens when you use phpstorms rebase, it copied the commits...

Used git push --force for the first time ever :boom:

swalkinshaw:
Thanks!

M13Kd:
@M13Kd

-------- Original message --------

PR #139, opened 2015-02-22
Fix reference to WordPress 4.1.1

emzo:
Bedrock was still referencing WordPress 4.1 in the composer.lock file. This updates the hash to point to version 4.1.1


swalkinshaw:
Good catch, thanks

PR #144, opened 2015-03-01
Autoloader: load plugins on muplugins_loaded

swalkinshaw:
Directly calling `loadPlugins` wasn't correct since all the autoloaded plugins would be loaded when the Autoloader was loaded. If another muplugin happened to be loaded before it, it wouldn't have access to the
rest of the plugins.

Using the `muplugins_loaded` action ensures consistency. All the autoloaded plugins will be guaranteed to be loaded at the same part of the loading process now and have access to all regular muplugins.

### DIFF
--- a/web/app/mu-plugins/bedrock-autoloader.php
+++ b/web/app/mu-plugins/bedrock-autoloader.php
@@ -2,7 +2,7 @@
 /**
  * Plugin Name: Bedrock Autoloader
  * Plugin URI: https://github.com/roots/bedrock/
- * Description: An autoloader that enables standard plugins to be required just like must-use plugins. The autoloaded plugins are included during mu-plugin loading. An asterisk (*) next to the name of the plugin designates the plugins that have been autoloaded.
+ * Description: An autoloader that enables standard plugins to be required just like must-use plugins. The autoloaded plugins are included during the `muplugins_loaded` action. An asterisk (*) next to the name of the plugin designates the plugins that have been autoloaded.
  * Version: 1.0.0
  * Author: Roots
  * Author URI: https://roots.io/
@@ -32,7 +32,7 @@ class Autoloader {
       add_filter('show_advanced_plugins', array($this, 'showInAdmin'), 0, 2); // Admin only filter.
     }
 
-    $this->loadPlugins();
+    add_action('muplugins_loaded', [$this, 'loadPlugins'], 0);
   }
 
   /**


aaemnnosttv:
I don't entirely agree with this change, primarily because `muplugins_loaded` may be too late in some cases.  Particularly, in a multisite environment, any network-enabled plugin would be loaded first, before the bedrock autoloaded mu-plugins ([source](https://github.com/WordPress/WordPress/blob/921ff2746dbed2a5234b5f80f766b39800c01367/wp-settings.php#L170-L190)).  This would mean that any _normal_ network-enabled plugin would be loaded before an autoloaded mu-plugin.  I see this potentially causing much more problems than the current implementation.

MU plugins should always load before any normal plugin (IMO), as that's one of the key defining features as mentioned in [the codex](http://codex.wordpress.org/Must_Use_Plugins)

If a hooked loading is desired for consistency - rather than the current immediate loading (which is more true to a real mu-plugin), I would recommend an alternative:

Add a new mu-plugin named something like `zz-bedrock-autoloader-hook.php`.  This plugin would contain the hook that the bedrock-autoloader would hook to for loading the plugins (eg: `bedrock_muplugins_loaded`)

Because mu-plugins are loaded in alphabetical order, the hook plugin would be loaded last, thus making all other mu-plugins available at autoloader-loaded load time, yet still loading before network-enabled or `muplugins_loaded`.

Alternatively, instead of adding a new plugin file for the hook, `bedrock-autoloader.php` could be renamed to `zz-bedrock-autoloader.php` and the hook could be called within the same file.

I would be happy to propose my own PR for this if others are open to considering it?

swalkinshaw:
@aaemnnosttv I get what you're saying and like usual we didn't really think of multisite. Maybe @Foxaii can chime in since this was his suggestion.

One of my problems with your proposal is the `zz` prefix naming. You get into this problem where someone could have a plugin that starts with `zzz` for example. Obviously rare, but it could happen.

Really annoying that WP doesn't have 2 distinct actions for muplugins and network ones.

aaemnnosttv:
I agree with the weakness of the prefix and the annoyance of the multisite weirdness.  I looked into it a bit more and rather than prefixing with `zz` it could use a tilde `~`, which should be a [cross-platform valid filename character](http://superuser.com/a/358861/150252), and will always be sorted/loaded after files that start with z.

As a test, I created a few simple mu-plugins (not bedrock autoloaded) to check the load order:
`app/mu-plugins/`
- a.php
- b.php
- c.php
- z.php
- ~last.php

Each plugin contained the same code to simply output the name of the file to indicate the order they were loaded in 
```php
<pre><?php var_dump( basename(__FILE__) ); ?></pre>
```
Here is what I got:

```
string(5) "a.php"
string(5) "b.php"
string(5) "c.php"
string(5) "z.php"
string(9) "~last.php"
```

Thoughts?

swalkinshaw:
It's definitely better than an alphabetical prefix. I'd like to confirm it works on Windows with no issue as well (assuming you didn't test on it).

I think regardless of the solution we need to have some more clear docs in the `README` showing the usual plugin load order and where the autoloader fits into that.

QWp6t:
Windows can have `~` in the file name. It's actually been used by Windows itself since at least the Windows 3.x days as well as Microsoft Office for just as long.

aaemnnosttv:
The tilde is and has been supported by Windows as a [filename character](http://en.wikipedia.org/wiki/Tilde#Microsoft_filenames) for a long time as @QWp6t said.  MS Word is one example that still uses it today as the first character in some of it's temporary files.  I can test on Windows, but we should probably have a few others test as well to be certain.

Foxaii:
In my opinion, because WordPress has an event-driven architecture, we should tie the auto loader to a standard event, instead of using an obscure filename and creating our own action as an ugly workaround to an uncommon issue.

If you must have an mu-plugin accompanied by a network plugin (or vice versa), why not have two network plugins? Surely both would be controlled the same way?

Can anyone give an example of a popular plugin that can only be modified by mu-plugins? Surely the vast majority of customisations are possible at theme level (i.e. via functions.php), which would make the mu/network plugin load order immaterial.

Ideally WordPress would fire `muplugins_loaded`, then have a `networkplugins_loaded` action, but it doesn't, unless someone wants to try their luck on trac.

Whatever is decided I'm happy to better document it.

francescolaffi:
I don't understand why its not correct to load the plugins as soon as the bedrock autoloader is included.

@swalkinshaw said
> If another muplugin happened to be loaded before it, it wouldn't have access to the
rest of the plugins.
I don't see this as a problem, when any plugin is loaded there could still be plugins waiting to be loaded, and since wp doesn't provide dependencies management its the plugin job to postpone its operations until all plugins are loaded if needed.

swalkinshaw:
@francescolaffi I seem to keep going back and forth on this, but I'm leaning towards what you said and keeping it as is. Obviously the autoloader is no different than the current plugin loading process right now.

QWp6t:
I agree that the whole alphabetical order thing sounds gimmicky and weird. The event hooks and prioritization exist for a reason.

swalkinshaw:
Keeping this as is.

PR #145, opened 2015-03-03
Switch to __DIR__ instead of dirname()

kix:
This should make the overall bedrock experience a bit faster. See [this StackOverflow thread](http://stackoverflow.com/questions/2749416/is-there-any-difference-between-dir-and-dirname-file-in-php) on why this is better.

As I see in `composer.json`, minimum PHP version supported is 5.4, so this should work OK.

### DIFF
--- a/config/application.php
+++ b/config/application.php
@@ -17,7 +17,7 @@ Dotenv::required(array('DB_NAME', 'DB_USER', 'DB_PASSWORD', 'WP_HOME', 'WP_SITEU
  */
 define('WP_ENV', getenv('WP_ENV') ? getenv('WP_ENV') : 'development');
 
-$env_config = dirname(__FILE__) . '/environments/' . WP_ENV . '.php';
+$env_config = __DIR__ . '/environments/' . WP_ENV . '.php';
 
 if (file_exists($env_config)) {
   require_once $env_config;
--- a/web/index.php
+++ b/web/index.php
@@ -2,4 +2,4 @@
 
 // WordPress view bootstrapper
 define('WP_USE_THEMES', true);
-require(dirname(__FILE__) . '/wp/wp-blog-header.php');
+require(__DIR__ . '/wp/wp-blog-header.php');


QWp6t:
:+1:

Just for reference in case anyone is wondering: `dirname()` doesn't actually care about the filesystem. It's just parsing a string. The performance difference between `__DIR__` and `dirname(__FILE__)` is negligible. They are the exact same thing. But since everyone using Bedrock has access to `__DIR__`, I agree it should be used.

kix:
@QWp6t, actually, there is a difference on the opcode level. Let's take a look:
```
<?php
echo dirname(__FILE__);
```
…transforms to these opcodes:
```
line     #* E I O op                           fetch          ext  return  operands
-------------------------------------------------------------------------------------
   2     0  E >   EXT_STMT
         1        EXT_FCALL_BEGIN
         2        SEND_VAL                                                 '%2FUsers%2Fkix%2FDocuments%2FCode%2Fphp%2Fopcode-test%2Fdirname.php'
         3        DO_FCALL                                      1  $0      'dirname'
         4        EXT_FCALL_END
         5        ECHO                                                     $0
   3     6      > RETURN                                                   1
```

And this:
```
<?php
echo __DIR__;
```
…looks like that in opcodes:
```
line     #* E I O op                           fetch          ext  return  operands
-------------------------------------------------------------------------------------
   2     0  E >   EXT_STMT
         1        ECHO                                                     '%2FUsers%2Fkix%2FDocuments%2FCode%2Fphp%2Fopcode-test'
   3     2      > RETURN                                                   1
```
As you can see, `dirname()` approach requires `FCALL`, while in case of `__DIR__` only a compile-time constant is being used.

This affects performance, but just a slightest bit (I've wrapped the `echo` call in a `for` loop that repeats 1 000 000 times):

```
$ time php const.php  > /dev/null
php const.php > /dev/null  0.77s user 0.56s system 60% cpu 2.221 total
```
```
$ time php dirname.php  > /dev/null
php dirname.php > /dev/null  1.86s user 0.50s system 99% cpu 2.362 total
```

swalkinshaw:
@kix there's another occurrence in `config/application.php` if you want to update that as well.

QWp6t:
> This affects performance, but just a slightest bit

Right, which is why I said...
> The performance difference between `__DIR__` and `dirname(__FILE__)` is negligible

swalkinshaw:
@kix ping

kix:
@swalkinshaw, yup, thanks for pointing it out. It's fixed now.

PR #150, opened 2015-03-15
Add dd() function to dev environment

QWp6t:
Laravel has a function called `dd()` which is short for "dump and die." This is my take on it for us WordPress users. I put it in `development.php` because it shouldn't really be used in any other setting.

### DIFF
--- a/config/environments/development.php
+++ b/config/environments/development.php
@@ -11,3 +11,13 @@ define('WP_SITEURL', getenv('WP_SITEURL'));
 define('SAVEQUERIES', true);
 define('WP_DEBUG', true);
 define('SCRIPT_DEBUG', true);
+
+if (!function_exists('dd')) {
+  function dd() {
+    $xdebug = ini_get('xdebug.overload_var_dump');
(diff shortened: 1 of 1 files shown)


QWp6t:
For anyone who's interested, I also use a `_log()` function that's similar but dumps to the error log.

```php
define('WP_DEBUG_LOG', true); // logs errors to web/app/debug.log
function _log() {
    $xdebug = ini_get('xdebug.overload_var_dump');
    $args = func_get_args();

    $xdebug && ini_set('xdebug.overload_var_dump', false);
    foreach ($args as $arg) {
        ob_start();
        var_dump($arg);
        error_log(trim(ob_get_clean()));
    }
    $xdebug && ini_set('xdebug.overload_var_dump', true);
}
```

swalkinshaw:
Just a note that we might be adding (or suggesting) via Composer a library/package which takes care of this. It's a better place to add this functionality vs in a config file.

mAAdhaTTah:
:+1:

PR #151, opened 2015-03-17
Added plugin meta for nicer output

rasmusbe:
Added plugin meta for nicer output in Must use plugin listing.


swalkinshaw:
Thanks for this. But we've since changed the URLs a bit.

Could you update them to `https://roots.io/bedrock/` and `https://roots.io`?

rasmusbe:
Done! :)

swalkinshaw:
Thanks!

PR #163, opened 2015-04-11
Move Capistrano configs into own repo

swalkinshaw:
The Capistrano configs are being moved into their own repo at https://github.com/roots/bedrock-capistrano.

This is being done for two reasons:

1. Cleaner Bedrock projects if you don't use Capistrano.
2. Better separation since people can be opinionated about deployment tools (especially Capistrano since it's a Ruby project).
3. [bedrock-ansible](https://github.com/roots/bedrock-ansible) now has Ansible based-deploys. See this [PR](https://github.com/roots/bedrock-ansible/pull/95)


M13Kd:
+m13kd@aol.com
+17705724576

PR #164, opened 2015-04-17
DRY up environment configs

swalkinshaw:
No point in repeating these settings in every env config since it's all the same, so we'll move them to `application.php`.

### DIFF
--- a/config/application.php
+++ b/config/application.php
@@ -24,6 +24,12 @@ if (file_exists($env_config)) {
 }
 
 /**
+ * URLs
+ */
+define('WP_HOME', getenv('WP_HOME'));
+define('WP_SITEURL', getenv('WP_SITEURL'));
+
+/**
  * Custom Content Directory
  */
 define('CONTENT_DIR', '/app');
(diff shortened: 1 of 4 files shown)


QWp6t:
I'd load environment php files first. Then conditionally define each of these, like so...

```php
 /**
 * URLs
 */
defined('WP_HOME') || define('WP_HOME', getenv('WP_HOME'));
defined('WP_SITEURL') || define('WP_SITEURL', getenv('WP_SITEURL'));
```

swalkinshaw:
@QWp6t environment files are loaded before anything else.

QWp6t:
Ahh, cool. My suggestion to conditionally define each constant still stands, though. I think it would be a good idea to let developers define stuff in their environment php files *if they want to*. They might need to apply some sort of logic to determine how the constants are defined in certain environments, but not in other environments.

In fact, I'd make that suggestion for every defined constant in application.php.


Up at the top of application.php, I suggest the following...
```php
<?php namespace Bedrock;

function define($name, $value) {
  defined($name) || \define($name, $value);
}

// ...
```

swalkinshaw:
Right now we just kind of punt on it and mention it in the `README`:

> Note: You can't re-define constants in PHP. So if you have a base setting in application.php and want to override it in production.php for example, you have a few options:

> Remove the base option and be sure to define it in every environment it's needed
Only define the constant in application.php if it isn't already defined.

Just not sure about namespacing the config and providing on our `define`.

Going to merge this as is for now and we can thin about your solution separately.

PR #169, opened 2015-04-23
Cleaner syntax

aaemnnosttv:
Couple of minor tweaks here in `application.php`.

Using short array syntax per style guidelines, as well as short ternary syntax for getenv defaults.

Checked with codesniffer.

### DIFF
--- a/config/application.php
+++ b/config/application.php
@@ -15,7 +15,7 @@ Dotenv::required(['DB_NAME', 'DB_USER', 'DB_PASSWORD', 'WP_HOME', 'WP_SITEURL'])
  * Set up our global environment constant and load its config first
  * Default: development
  */
-define('WP_ENV', getenv('WP_ENV') ? getenv('WP_ENV') : 'development');
+define('WP_ENV', getenv('WP_ENV') ?: 'development');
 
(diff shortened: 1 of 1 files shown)


swalkinshaw:
Thanks!

config/application.php
```diff
@@ -9,13 +9,13 @@
...
 
-Dotenv::required(array('DB_NAME', 'DB_USER', 'DB_PASSWORD', 'WP_HOME', 'WP_SITEURL'));
+Dotenv::required(['DB_NAME', 'DB_USER', 'DB_PASSWORD', 'WP_HOME', 'WP_SITEURL']);
 
 /**
  * Set up our global environment constant and load its config first
  * Default: development
  */
-define('WP_ENV', getenv('WP_ENV') ? getenv('WP_ENV') : 'development');
+define('WP_ENV', getenv('WP_ENV') ?: 'development');
```
mAAdhaTTah: This syntax works...? Didn't know that!

swalkinshaw: http://php.net/manual/en/language.operators.comparison.php#language.operators.comparison.ternary

PR #178, opened 2015-05-11
Adding dependency badge to README

reiz:


swalkinshaw:
@reiz thanks but I had to remove this for now. The badge takes forever to load and the GitHub proxy just times out resulting in a broken image.

PR #183, opened 2015-05-27
PSR-0 Deprecation: use PSR-4

KLVTZ:
PSR-0 is deprecated:
https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-0.md

We should use PSR-4, the recommended alternative.


swalkinshaw:
Thanks!

PR #184, opened 2015-05-28
Remove installation script

swalkinshaw:
This removes the installer script which would optionally create a
`.env` with generated WordPress salts/keys.

If Bedrock is being used for development/local purposes then the salts
aren't even needed. WordPress will automatically generate them for you.

For staging/production/remove environments, salts should be managed
through proper configuration management (such as what bedrock-ansible
does).


etcook:
@swalkinshaw I used this to sometimes generate random salts for production environments. What if we just left the script there, but allowed it only to be manually invoked?

swalkinshaw:
@etcook yeah unfortunately https://api.wordpress.org/secret-key/1.1/salt/ this page generates the PHP code. We should probably just create a simple Composer package or WP CLI plugin or something to generate these. I think it's just useful just don't think it belongs in Bedrock.

rossedman:
:+1: Separation of concerns

swalkinshaw:
Found this: https://github.com/sebastiaandegeus/wp-cli-salts-command/pull/3

rossedman:
@swalkinshaw I think that's a good solution. If someone wanted to trigger this after a composer build they could just use composer post-install

aaemnnosttv:
Enter the [`dotenv` command](https://github.com/aaemnnosttv/wp-cli-dotenv-command)

swalkinshaw:
@aaemnnosttv awesome :+1:

PR #192, opened 2015-07-04
Add .DS_Store to .gitignore

duncanjbrown:
This PR adds `.DS_Store` to `.gitignore` so these useless files that OS X generates don't get checked in.


swalkinshaw:
Rules for files like this belong in your own personal global gitignore. See https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

PR #193, opened 2015-07-10
Only call required if .env file exists

pcfreak30:
Fixes a crash in Dotenv required which expects a valid Loader instance in $this->loader in `vendor/vlucas/phpdotenv/src/Dotenv.php` on line `83`

### DIFF
--- a/config/application.php
+++ b/config/application.php
@@ -8,9 +8,8 @@ $webroot_dir = $root_dir . '/web';
 $dotenv = new Dotenv\Dotenv($root_dir);
 if (file_exists($root_dir . '/.env')) {
   $dotenv->load();
+  $dotenv->required(['DB_NAME', 'DB_USER', 'DB_PASSWORD', 'WP_HOME', 'WP_SITEURL']);
 }
-
-$dotenv->required(['DB_NAME', 'DB_USER', 'DB_PASSWORD', 'WP_HOME', 'WP_SITEURL']);
 
 /**
  * Set up our global environment constant and load its config first


austinpray:
[yup](https://github.com/vlucas/phpdotenv/blob/1f207799d61a7d0fc50dfcedbe0ecf350a770af4/src/Dotenv.php#L83) and [yup](https://github.com/vlucas/phpdotenv/blob/1f207799d61a7d0fc50dfcedbe0ecf350a770af4/src/Loader.php#L51). :+1:

austinpray:
@louim looks like [it's all your fault](https://github.com/roots/bedrock/commit/2281dbe865be7ad9a034829cdeaff93616ee3a1b).

QWp6t:
Minus whale put `$dotenv = new Dotenv\Dotenv($root_dir);` in there too since it's not used anywhere else in the script. Just keep it all contained in its own conditional statement.

louim:
@austinpray not totally my fault :P. In v1.x dotenv would validate even when not loading file. See :https://github.com/vlucas/phpdotenv/issues/101

Bedrock doesn't care (and I think it is a good idea) from where the ENV vars  come from ( could be passed as server vars, or directly loaded from `$ENV=...`). 

But it will check if the variables that are needed are set , hence the `required` that is called outside the condition. I see that it is a problem in the current dotenv version, and we should find a temporary solution until it is fixed. @pcfreak30 have you checked if dotenv throws an error if you remove the condition and don't provide a `.env` file in the bedrock folder?

swalkinshaw:
@louim I agree with your point but I'm not sure that Dotenv should be responsible for enforcing general ENV vars. After all the library is specifically for loading `.env` files. If a `.env` file doesn't exist, should Dotenv even be used for anything else?

:+1: to @QWp6t's suggestion as well.

louim:
@swalkinshaw  I tried (and failed) to keep the original behaviour of Bedrock. I think that since we require variable to be loaded from the environment (be it through Dotenv or another mean), and that Dotenv provide a handy method to check that the variables are set (it does more that just checking if the var is defined) we could keep using that. But it would require the upstream to be fixed to work like it was in v1.

Either way, I fine with both options.

pcfreak30:
I am using this while creating a docker stack, so ENV is passed via docker-compose and shell entry script, thus how I caught the errors. No ENV used.

See https://github.com/pcfreak30/bedrock-docker :).

swalkinshaw:
Merging this for now until we get a better fix.

louim:
Fixes for this issue have been merged upstream https://github.com/vlucas/phpdotenv/pull/102. Ill keep an eye on this when it land on a stable release.

pcfreak30:
Any ETA for tagging a new release? Would like to update my docker repo back to using the official git repo.

louim:
@pcfreak30 your fix is merged, so you can use the official repo :smile: I'll open another pull request when  the fix I made upstream is ready to be implemented in Bedrock.

pcfreak30:
@louim I know, but my Dockerfile pulls the latest tags aka release rather than the head of the repo to ensure its a stable version.

PR #201, opened 2015-09-03
Bump minimum required PHP version to 5.5

retlehs:
PHP 5.4 EOL is in 10 days: http://php.net/supported-versions.php


austinpray:
Midnight merge party on the 14th at my place

swalkinshaw:
party right now

PR #206, opened 2015-10-12
Set DISABLE_WP_CRON from env var

swalkinshaw:
This allows better control via configuration management like Trellis.

Also change the default to `false`. This setting requires external server configuration for cron to work and since Bedrock is a standalone boilerplate, we shouldn't require that by default.


retlehs:
docs will need updating: https://roots.io/bedrock/docs/wp-cron/

swalkinshaw:
@retlehs no need for that doc anymore

PR #211, opened 2015-10-23
Remove unnecessary slash from $root_dir

damianwajer:
Fix for environments where `dirname(__DIR__)` returns just a slash `/` which makes `$webroot_dir` = `//web`.

A real life example with which I had to deal recently:

Plugin defines its URL by creating a path:

```
$plugin_dir = dirname(__FILE__); // /web/app/plugins/example-plugin
```

Then it creates an URL like this:

```
$plugin_url = str_replace(
	array( WP_CONTENT_DIR, WP_PLUGIN_DIR ),
	array( WP_CONTENT_URL, WP_PLUGIN_URL ),
	$plugin_dir;
);
```

What we have now is:

`$plugin_dir` = `/web/app/plugins/example-plugin`

`WP_CONTENT_DIR` = `//web/app`,
 `WP_PLUGIN_DIR` = `//web/app/plugins`

`WP_CONTENT_URL` = `http://www.example.com/app`, 
`WP_PLUGIN_URL` = `http://www.example.com/app/plugins`

In this case `str_replace` doesn't work as expected, URL doesn't exist: 
`$plugin_url` = `/web/app/plugins/example-plugin/`.

It would work with fixed $root_dir:

`$plugin_dir` = `/web/app/plugins/example-plugin`

`WP_CONTENT_DIR` = `/web/app`, 
`WP_PLUGIN_DIR` = `/web/app/plugins`

`WP_CONTENT_URL` = `http://www.example.com/app`, 
`WP_PLUGIN_URL` = `http://www.example.com/app/plugins`

`/web/app/plugins/example-plugin` -> `http://www.example.com/app/plugins/example-plugin`

### DIFF
--- a/config/application.php
+++ b/config/application.php
@@ -1,5 +1,5 @@
 <?php
-$root_dir = dirname(__DIR__);
+$root_dir = rtrim(dirname(__DIR__), '/');
 $webroot_dir = $root_dir . '/web';
 
 /**


mAAdhaTTah:
> Fix for environments where dirname(__DIR__) returns just a slash / which makes $webroot_dir = //web.

This happens? I guess if you're serving from `/www/` or something.

Also, wouldn't this change always remove a leading `/`?

QWp6t:
@mAAdhaTTah
No because it's `rtrim()`. It will only remove trailing slash, not leading slash. `dirname()` doesn't have a trailing slash anyway, so this is fine in that regard.

@damianwajer 
I think you should go with this instead...
```diff
 $root_dir = dirname(__DIR__);
-$webroot_dir = $root_dir . '/web';
+$webroot_dir = rtrim($root_dir, '/') . '/web';
```

That way `$root_dir` is actually the root directory, i.e., `/`.

But I don't think this should be added to Bedrock by default. That's just too weird of a structure putting everything in the root of your filesystem like that. I think you should move your document root instead.

Or are there situations in which the directory structure might appear like this to PHP for security reasons or something? Like a form of masking the server's folder structure? Because ostensibly it looks like your directory root is `/web`, which is ... odd.

damianwajer:
I know that this kind of structure is weird, but apparently it happens sometimes in environments where domain directory `public_html` appears to PHP like the root directory `/` (probably like you've said for security reasons - a form of masking the server's folder structure). 

In order to make Bedrock works, I had to point a domain to a subdirectory `/web` (which in my case is `/public_html/web` but appears to PHP as `/web`). Maybe the better solution is to create another directory, e.g. `/www` move Bedrock there, then point a domain to a subdirectory `/www/web` so document root is `/www/web` and `$root_dir` is `/www`.

I'm not sure if it should be added to Bedrock by default but it's an issue that may occurs in some shared hosting environments.

Anyway, your solution will also work.

QWp6t:
Looks like we're gonna pass on this one for now. We might revisit it if at some point if there's a compelling reason to do so.

PR #214, opened 2015-10-27
In the autoloader, remove argument from call to get_mu_plugins()

bendoh:
`get_mu_plugins()` is being called with an argument, but this function does not accept one. I checked back to 3.0.0 when it was introduced and it doesn't appear this API call ever accepted an argument.

This PR simply removes the argument from the call for clarity's sake.


swalkinshaw:
Good catch. Thanks :)

PR #233, opened 2016-01-27
Use oscarotero/env instead of `getenv`

QWp6t:
See https://github.com/roots/bedrock/pull/229 for background.

The `env` function from https://github.com/oscarotero/env handles simple coercions for values like 'True' (`true`) which is needed.


kalenjohnson:
I approve of this

PR #239, opened 2016-02-16
Add BEDROCK_SECURE_ENV env variable

jamesgol:
If set to true then all of the secure variables are unset after they are used for configuring WordPress
Disabled by default, so it must be enabled to have any effect

I really like bedrock, but I dislike the idea of using environment variables for secure things. I've seen far too many plugins that allow unrestricted access to the phpinfo() function which would give enough information away for someone to completely takeover a site.

### DIFF
--- a/config/application.php
+++ b/config/application.php
@@ -71,6 +71,29 @@ define('AUTOMATIC_UPDATER_DISABLED', true);
 define('DISABLE_WP_CRON', env('DISABLE_WP_CRON') ?: false);
 define('DISALLOW_FILE_EDIT', true);
 
+if (true === env('BEDROCK_SECURE_ENV')) {
+  // If BEDROCK_SECURE_ENV is set to true, then unset all the security sensitive variables after they are used
+  bedrock_clear_env('DB_NAME');
+  bedrock_clear_env('DB_USER');
+  bedrock_clear_env('DB_PASSWORD');
+  bedrock_clear_env('DB_HOST');
+  bedrock_clear_env('DB_PREFIX');
+  bedrock_clear_env('AUTH_KEY');
+  bedrock_clear_env('SECURE_AUTH_KEY');
+  bedrock_clear_env('LOGGED_IN_KEY');
+  bedrock_clear_env('NONCE_KEY');
+  bedrock_clear_env('AUTH_SALT');
+  bedrock_clear_env('SECURE_AUTH_SALT');
+  bedrock_clear_env('LOGGED_IN_SALT');
+  bedrock_clear_env('NONCE_SALT');
+}
+
+function bedrock_clear_env($var_name) {
+  // Dotenv has a similar function Dotenv\Loader\clearEnvironmentVariable() but it is not exposed
+  putenv($var_name);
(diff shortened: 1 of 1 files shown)


kalenjohnson:
I don't think this is a bad idea. Are there any main plugins that actually include `phpinfo()` in their code though? I don't recall seeing this. Scary if that's the case though

jamesgol:
I just searched through plugins I have on a development system and there are 2 plugins that call it, but those are each calling it from a secure location.  I have definitely seen it exposed before only plugins that I've audited for security issues.

QWp6t:
In my opinion, this shouldn't be optional. It should just be done. Also, [we can have dotenv do it](https://github.com/vlucas/phpdotenv/blob/master/src/Loader.php#L344-L370).

jamesgol:
I was originally just going to submit a PR that would just do it, but figured it would get more traction by having it be configurable.

I would have used the dotenv function if it was exposed, but alas it is not exposed because $loader is protected:
```
    /**
     * The loader instance.
     *
     * @var \Dotenv\Loader|null
     */
    protected $loader;
```

QWp6t:
Ahh, I see. That makes sense. What an odd design, to have a buried "utility method" that can't even be accessed.

jamesgol:
The design is a bit odd to me also. It would make sense if \Dotenv\Loader\load() returned a self reference, but it returns an array of all the lines.

jamesgol:
Another possibility to doing this instead of adding a function would be to just look using an array
```
foreach (array( 'DB_NAME', 'DB_USER', ... ) as $key) {
  putenv($key);
  unset($_ENV[$key], $_SERVER[$key]);
}
```

QWp6t:
So far, we like
```php
array_map('putenv', array_keys(get_defined_constants(true)['user']));
```
But we might wind up doing something different instead. We're gonna keep chatting about it and figure out how to proceed.

jamesgol:
Unfortunately putenv doesn't covert it completely, since the keys are also in $_ENV and $_SERVER.

QWp6t:
That's fine. We're not concerned about the existence of the keys in the superglobals. We're actually concerned about the environment variables being leaked to child processes and whatnot. Their presence in `phpinfo()` is inconsequential considering that the the global constants are all there anyway.

jamesgol:
Being leaked via phpinfo() was/is my biggest concern, because I have seen plugins give open access to that function, though I was not using Bedrock (or anything that stored important data in env vars) at the time so it wasn't a major deal. I worked with the developers to restrict access to that, but if it's happened before it will happen again.

swalkinshaw:
@jamesgol thanks for bringing this up. It led to us talking about our current usage of Dotenv and we realized that we should move away from it.

The `.env` file has one main purpose: it keeps your "secrets" in a separate file out of version control. But this comes with downsides: 1) exposes things to `phpinfo`, and more importantly, 2) exposes the env vars to any child processes.

In our case, there's no actual reason to use real environment variables. We just need a simpler way of setting variables for use in config files.

We're pretty sure we're going to switch to an `ini` file in the future since PHP has `parse_init` built-in. Because of this, I'm going to close this PR.

Issue #241, opened 2016-02-29
Composer transportException


peebeebee:
Problem as seen in https://github.com/outlandishideas/wpackagist/issues/138  .
In the meantime we could add the following to the composer config?

```
{
        "config": {
            "secure-http": false
        }
   }
```

PR #242, opened 2016-02-29
Updated composer to allow non-https connections

peebeebee:
WPackagist can not use https-connections right now. 
In the mean time, this could be a fix.

Cfr: https://github.com/roots/bedrock/issues/241

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,8 @@
     "forum": "https://discourse.roots.io/category/bedrock"
   },
   "config": {
-    "preferred-install": "dist"
+    "preferred-install": "dist",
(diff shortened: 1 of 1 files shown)


swalkinshaw:
Thanks for bringing this up. It's been fixed upstream though in https://github.com/outlandishideas/wpackagist/issues/138. And fixed in Bedrock by https://github.com/roots/bedrock/commit/48999fe6f452b57a7012444811aa1a9fc13e6669

singlow:
Does updating the wpackagist repo url to https fix this for others? I am still getting non-https svn urls from wpackagist which cause the same exception.

PR #245, opened 2016-03-24
Instruct users to quote salts in .env.example

bjorn-ali-goransson:
If we use the Salt function on api.wordpress.org, the keys will usually contain spaces. To avoid the `'Dotenv\Exception\InvalidFileException' with message 'Dotenv values containing spaces must be surrounded by quotes.'`, we should instruct users to quote their keys.


swalkinshaw:
:+1: thanks.

PR #249, opened 2016-04-13
Update salt link

vinkla:
Replaced the default salt link with a one that works better for dotenv files. Now developers just need to copy and paste the environment salt keys to their `.env` file.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,7 +7,7 @@ WP_ENV=development
 WP_HOME=http://example.com
 WP_SITEURL=${WP_HOME}/wp
 
-# Generate your keys here (wrap them in quotes): https://api.wordpress.org/secret-key/1.1/salt/
+# Generate your keys here (wrap them in quotes): https://wordplate.github.io/salt/
 AUTH_KEY='generateme'
 SECURE_AUTH_KEY='generateme'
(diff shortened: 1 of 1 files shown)


swalkinshaw:
I agree it's nicer but that salt generator should probably not be used (for production purposes). It uses a [JavaScript function](https://github.com/wordplate/wordplate.github.io/blob/044735ce0561c79d56a7956e74fae8bf5eb8d386/app/salt/index.html#L25-L44) and `Math.random` which is not cryptographically secure. See http://stackoverflow.com/questions/5651789/is-math-random-cryptographically-secure for example.

austinpray:
https://jsfiddle.net/austinpray/f1uu2st7/9/ -> roots.io/salt.html

vinkla:
@swalkinshaw okay, I understand. Do you know how WordPress handles this on their salt page?

swalkinshaw:
We've updated this to an improved version on roots.io in https://github.com/roots/bedrock/commit/43ff1ed14706a5af0c486d5ab6356f40df8b1589.

@vinkla I have no idea what the WP page uses. I'm sure it's done in PHP.

DavidSchargel:
Why do simple things like this make me so happy? Thanks, @austinpray!

PR #251, opened 2016-04-14
Fix URL issues on multisite (especially in network admin)

### LINKED ISSUE #250: URL issues on multisite
I have recently set up my first multisite with Bedrock, and I have encountered several issues where the URLs are not correct. After some investigation, I figured that the issues are a result of a problem in WordPress Core - in some places it does not honor the site URL setting. This is no problem with a regular WP install, but when WordPress is installed in a subdirectory as it is in Bedrock, the site URL is different from the home URL.

The most significant problem is that some areas in the network admin will direct me to /wp-admin instead of /wp/wp-admin, for example when you're on the My Sites section and click on "Edit Site".

To be clear, this is not a Bedrock bug, but I think it would be great if Bedrock provided a fix for it, so that other people do not run into this problem. While there are ways to workaround this with nginx / htaccess rewrites, I'd prefer to actually fix it from the code site. As of now, I have fixed these issues on my site, so I'll create a pull-request with my approach.

Btw I also created a ticket on Core Trac, the first responses told me that WP doesn't officially support this - so not sure whether there will be a Core fix soon.

felixarntz:
This fixes #250

### DIFF
--- a/web/app/mu-plugins/ms-url-fixer.php
+++ b/web/app/mu-plugins/ms-url-fixer.php
@@ -15,6 +15,12 @@ if ( ! is_multisite() ) {
     return;
 }
 
+/**
+ * Class URLFixer
+ * @package Roots\Bedrock
+ * @author Roots
+ * @link https://roots.io/
+ */
 class URLFixer {
     /** @var Roots\Bedrock\URLFixer Singleton instance */
     private static $instance = null;
--- a/web/app/mu-plugins/ms-url-fixer.php
+++ b/web/app/mu-plugins/ms-url-fixer.php
@@ -1,0 +1,89 @@
+<?php
+/**
+ * Plugin Name: Multisite URL Fixer
+ * Plugin URI: https://github.com/roots/bedrock/
+ * Description: Fixes WordPress issues with home and site URL on multisite.
+ * Version: 1.0.0
+ * Author: Roots
+ * Author URI: https://roots.io/
+ * License: MIT License
+ */
+
+namespace Roots\Bedrock;
+
+if ( ! is_multisite() ) {
+    return;
+}
+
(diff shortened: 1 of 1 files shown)


swalkinshaw:
Thanks for sharing this @felixarntz. Just wanted to reply that I'm not sure if we want to provide extra code in Bedrock to fix this yet.

I'll leave this open for the time being but it would also need testing and hopefully other people who have more Multisite experience than me can chime in as well.

reverbsoul:
Thanks for this @felixarntz - been looking for this fix for a while.

oxyc:
No extensive testing done but so far so good–all the URLs seem to be fixed with this PR.

xyNNN:
What's the state of this PR?

swalkinshaw:
I don't have enough multisite experience to merge this without knowing the implications yet.

But this PR also won't be merged regardless until it passes our coding standards (see CI failures for details).

cfxd:
Hey @felixarntz thanks for this PR. I'd love to see this pass build checks. That would be one step closer to a merge and fix of this issue which affects both Bedrock and Trellis (https://github.com/roots/trellis/pull/616).

swalkinshaw:
I think this should be a separate plugin/package. Although the problem happens with Bedrock, it's a WP but not a Bedrock one.

If this was a standalone package, Bedrock could require or at least recommend it.

amenk:
@felixarntz did this fix all issues? are you working on a standalone package?

lekkerduidelijk:
The PR #276 by @ryanplasma fixed my issues. It's a small mu-plugin file. 👏

JulienMelissas:
Just for reference purposes, https://github.com/roots/multisite-url-fixer should now help with this.

Issue #230, opened 2016-01-17
Use DB_CHARSET utf8mb4 by default


ideadapt:
Since April 2015, WordPress uses utf8mb4 by default. It would be nice if you do so as well in config/application.php. 
When I got this error using `wp import export.sql`:

`ERROR 1366 (HY000) at line 215: Incorrect string value: '\xF0\x9F\x8C\xAE" ...' for column 'option_value' at row 1`

It cost me some time to realize, that my wp-config was wrong. 

Changing 
`define('DB_CHARSET', 'utf8');`
to
 `define('DB_CHARSET', 'utf8mb4');`
resolved it.

swalkinshaw:
@ideadapt thanks, updated.

PR #256, opened 2016-05-15
Allow DB_CHARSET to be set in environment

yeah:
We'll still use utf8mb4 as a default if DB_CHARSET is undefined, but we allow users to define their own charset. Sadly, some MySQL installations out there in the wild still don't support utf8mb4...

### DIFF
--- a/config/application.php
+++ b/config/application.php
@@ -52,7 +52,7 @@ define('DB_NAME', env('DB_NAME'));
 define('DB_USER', env('DB_USER'));
 define('DB_PASSWORD', env('DB_PASSWORD'));
 define('DB_HOST', env('DB_HOST') ?: 'localhost');
-define('DB_CHARSET', 'utf8mb4');
+define('DB_CHARSET', env('DB_CHARSET') ?: 'utf8mb4');
 define('DB_COLLATE', '');
 $table_prefix = env('DB_PREFIX') ?: 'wp_';
 


yeah:
#230 breaks fresh installs on MySQL servers which don't support utf8mb4 yet.

Steps to reproduce:

1. Set up a MySQL server that doesn't support utf8mb4
2. Follow instructions at https://roots.io/bedrock/docs/installing-bedrock/
3. Complete installation wizard at `/wp/wp-admin`
4. An error message appears: `WordPress database error installation: Unknown character set: 'utf8mb4'`

Expected behavior: 

- WordPress should install cleanly.

This pull request allows users to define `DB_CHARSET=utf8` (without the `mb4`) in their `.env` to make things work.

austinpray:
Surely you should be using the same version of MySQL and the same DB charset between environments, right?

yeah:
I am not sure how that relates to the problem I mentioned. I'm new to bedrock, so please bear with me. How should I properly address the problem described above if not by this PR?

austinpray:
Environment variables are for settings that change between environments. If your production MySQL server is a version such that it doesn't support utfmb8 then you should change this config so it takes place in all environments.

yeah:
So, where exactly do you expect people to set the charset if they're running a MySQL version that doesn't support it? Am I expected to run a fork of bedrock in that case?

The point I'm trying to make is that https://roots.io/bedrock/docs/installing-bedrock/ does not result in a working WordPress install in case the MySQL server used doesn't support utf8mb4.

swalkinshaw:
You edit [this line](https://github.com/roots/bedrock/blob/303e59477bae734774bfeb17ccd19d94da622b6a/config/application.php#L55) or just remove it.

yeah:
Alright then. Should not https://roots.io/bedrock/docs/installing-bedrock/ mention a minimum requirement of MySQL 5.5.3 then?

yeah:
> edit: you should basically always be running a "fork" of Bedrock. It's a boilerplate project that you should customize.

ok, thanks.

PR #261, opened 2016-06-08
Add DB_PREFIX variable to .env.example

thedamon:
Followed format of php.ini where things that you can customize but often don't need to are present in the config file but commented out. This gives much quicker insight into what changes are possible.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,6 +7,9 @@ WP_ENV=development
 WP_HOME=http://example.com
 WP_SITEURL=${WP_HOME}/wp
 
+# Optional variables
+# DB_PREFIX=wp_
+
 # Generate your keys here: https://roots.io/salts.html
 AUTH_KEY='generateme'
 SECURE_AUTH_KEY='generateme'
--- a/.env.example
+++ b/.env.example
@@ -1,14 +1,14 @@
 DB_NAME=database_name
 DB_USER=database_user
 DB_PASSWORD=database_password
-DB_HOST=database_host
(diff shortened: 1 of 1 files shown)


.env.example
```diff
@@ -7,6 +7,9 @@ WP_ENV=development
 WP_HOME=http://example.com
 WP_SITEURL=${WP_HOME}/wp
 
+# Optional variables
```
swalkinshaw: Can you move this up under the rest of the DB vars?

Also maybe we should treat `DB_HOST` as an optional comment too. Its default is `localhost`.

thedamon: Interesting idea, although I feel like almost every case it will need a port number added, so I'm not sure the implication that it might be ok to leave it will help ppl. Although that may just be my own experience

swalkinshaw: I think the default port would be the most popular. Changing `DB_HOST` more than likely means you're just using a remote host (still with the default port).

Anyway, port has nothing to do with this. We should move `DB_HOST` to be an optional comment because it is literally optional. It has a default fallback value.

QWp6t:
Can you make the following change?

```diff
 DB_PASSWORD=database_password
-DB_HOST=database_host 
 
 # Optional variables
+# DB_HOST=localhost
 # DB_PREFIX=wp_
 
 WP_ENV=development
```

QWp6t:
@thedamon You wanna make the change noted above? Or should we handle this one on our own?

PR #265, opened 2016-06-29
Ignore irrelevant files when installing via Composer

shaneog:
When composer installs a stable version of a project it will use the
`dist` as the default.

This `.gitattributes` file excludes several files and directories from
being exported to the dist zip file.

This should make it easier for users to get started, since they don't
have to "clean up" the generated directory.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+/.github export-ignore
+/.gitattributes export-ignore
+/.editorconfig export-ignore
+/.travis.yml export-ignore
+/phpcs.xml export-ignore


shaneog:
Other files such as `.travis.yml`, `CHANGELOG.md`, `.gitignore`, etc. can also be removed from the `dist` export using this method.

swalkinshaw:
Do you want to add:

* `.editorconfig`
* `phpcs.xml`
* `.travis.yml`
* `.env.example`

Those should all be excluded. CHANGELOG, README, and LICENSE should stay.

shaneog:
Is `.env.example` not useful for new users? And copying it is mentioned in the README.

swalkinshaw:
@shaneog good point, let's keep it

shaneog:
@swalkinshaw Updated.

swalkinshaw:
Thanks!

PR #266, opened 2016-07-04
Add Gitlab CI tests

DanielDent:
This runs the same tests suite as the existing .travis.yml file. Users who fork this repo to a Gitlab server will be able to continue to use CI on their private fork. 

If desired, it would be easy to setup a mirror of the project on gitlab.com - that would make it so that the README could include a Gitlab build badge example.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,28 @@
+
+variables:
+  COMPOSER_HOME: composer-home
+
+cache:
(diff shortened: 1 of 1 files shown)


swalkinshaw:
@DanielDent I think this is something that belongs on the Wiki instead. We don't want to include files for people who *might* fork it and run tests on another CI service.

Feel free to create a Wiki page ("CI configuration" or something) if you want.

PR #268, opened 2016-07-13
Updated .gitignore

sergeyklay:
The `bin` is covered by `vendor/*`. Correct template is `bin/`


swalkinshaw:
I don't really understand this change. `bin` without the trailing slash properly ignores the entire `bin` directory in the root of the repo.

If I'm missing something please let me know.

PR #269, opened 2016-07-15
Update instructions for generating salts

ssteinerx:
This is just a small doc improvement:

* Show more explicit usage of `composer create-project` command
* Show more detailed usage and installation of aaemnnosttv/wp-cli-dotenv-command
* Abbreviated a couple of links  to make the text easier to edit

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ See a complete working example in the [roots-example-project.com repo](https://g
 
   `composer create-project roots/bedrock your-project-folder-name`
 
-1. Copy `.env.example` to `.env` and update environment variables:
+2. Copy `.env.example` to `.env` and update environment variables:
   * `DB_NAME` - Database name
   * `DB_USER` - Database user
   * `DB_PASSWORD` - Database password
@@ -51,9 +51,11 @@ See a complete working example in the [roots-example-project.com repo](https://g
 
   Or, you can cut and paste from the [Roots WordPress Salt Generator][roots-wp-salt].
 
-1. Add theme(s) in `web/app/themes` as you would for a normal WordPress site.
-1. Set your site vhost document root to `/path/to/site/web/` (`/path/to/site/current/web/` if using deploys)
-1. Access WP admin at `http://example.com/wp/wp-admin`
+3. Add theme(s) in `web/app/themes` as you would for a normal WordPress site.
+
+4. Set your site vhost document root to `/path/to/site/web/` (`/path/to/site/current/web/` if using deploys)
+
+5. Access WP admin at `http://example.com/wp/wp-admin`
 
 ## Deploys
 
--- a/README.md
+++ b/README.md
@@ -29,8 +29,11 @@ See a complete working example in the [roots-example-project.com repo](https://g
 
 ## Installation
 
-1. Create a new project - `composer create-project roots/bedrock`
-2. Copy `.env.example` to `.env` and update environment variables:
+1. Create a new project in a new folder for your project:
+
+  `composer create-project roots/bedrock your-project-folder-name`
+
+1. Copy `.env.example` to `.env` and update environment variables:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,19 +29,31 @@ See a complete working example in the [roots-example-project.com repo](https://g
...
-3. Add theme(s) in `web/app/themes` as you would for a normal WordPress site.
-4. Set your site vhost document root to `/path/to/site/web/` (`/path/to/site/current/web/` if using deploys)
-5. Access WP admin at `http://example.com/wp/wp-admin`
+  * `AUTH_KEY`, `SECURE_AUTH_KEY`, `LOGGED_IN_KEY`, `NONCE_KEY`, `AUTH_SALT`, `SECURE_AUTH_SALT`, `LOGGED_IN_SALT`, `NONCE_SALT`
+
+  If you want to automatically generate the security keys (assuming you have wp-cli installed locally) you can use the very handy [wp-cli-dotenv-command][wp-cli-dotenv]:
+
+      `wp package install aaemnnosttv/wp-cli-dotenv-command`
+
+      `wp dotenv salts regenerate`
```
swalkinshaw: the backticks on these 2 commands aren't needed, it's already in a code block

ssteinerx: Fixed, was previously inline, forgot to remove when I separated them.

README.md
```diff
@@ -29,19 +29,31 @@ See a complete working example in the [roots-example-project.com repo](https://g
...
+
+      wp package install aaemnnosttv/wp-cli-dotenv-command
+
+      wp dotenv salts regenerate
+
+  Or, you can cut and paste from the [Roots WordPress Salt Generator][roots-wp-salt].
+
+1. Add theme(s) in `web/app/themes` as you would for a normal WordPress site.
+1. Set your site vhost document root to `/path/to/site/web/` (`/path/to/site/current/web/` if using deploys)
+1. Access WP admin at `http://example.com/wp/wp-admin`
```
retlehs: replace `1.` + `1.` + `1.` etc with correct numbers

ssteinerx: This is the recommended way of doing numbered lists as Markdown will take care of renumbering them so, when you add an item, you don't have to manually redo everything.

ssteinerx: [Original Markdown Docs](https://daringfireball.net/projects/markdown/syntax)

Go there and search for: 

    If you instead wrote the list in Markdown like this:

There is a note there not to use the number 1. due to future versions starting at arbitrary numbers, but that hasn't happened yet.  If you'd like, I can change them all to '2.' but the way I did  it is fundamentally correct and makes future additions/subtractions from the list much less error-prone.

QWp6t: I think it should still be 1., 2., 3., etc. Otherwise it looks stupid when it's not rendered. The note regarding the use of `1.` is strictly for those who are using the "lazy list numbering."

> The point is, if you want to, **you can use ordinal numbers in your ordered Markdown lists, so that the numbers in your source match the numbers in your published HTML.** But **if you want to be lazy,** you don’t have to.

Sweet, we can choose between being lazy or not being lazy. Doesn't sound like either one is "recommended" or "instructed."

> **If you do use lazy list numbering,** however, you should still start the list with the number 1. At some point in the future, Markdown may support starting ordered lists at an arbitrary number.

This isn't an issue we need to worry about if we're not lazy.

So let's not be lazy.

ssteinerx:
Changed to un-lazy numbering, ready to merge.

retlehs:
torn on adding things to the readme & i've spent time on most of our projects making them as simple as possible while moving things into docs.

this PR could be moved to https://github.com/roots/docs/blob/master/bedrock/installing-bedrock.md

swalkinshaw:
Thanks, we can always update this in the docs as well.

PR #270, opened 2016-08-06
Modular configuration

aaemnnosttv:
This PR splits Bedrock's application.php into it's logical modular components of configuration, aimed to make Bedrock more extensible and easier to maintain.

Modeled after conventions seen in *nix `conf.d` type directories, configuration modules are just php files, named in a way which controls the load order, similar to action/filter priorities.

image

This makes it so modules of configuration are more reusable between projects, as well as making it easier for sites to "sync" with future changes to Bedrock core.  This has been difficult at times on projects with a customized `application.php`.

For example, I often add my own logging configuration to the application.php.  With a modular configuration, this could be as simple as adding a `config/modules/0-logging.php` file; no need to change anything related to Bedrock core.

An important change which makes this PR possible, is the introduction of new constants `BEDROCK_ROOT_DIR` and `BEDROCK_WEB_ROOT_DIR`, which not only replace the need for their respective temporary variables in the various configuration modules, but can be useful in other cases with your own application code.

Eg: Loading a file that is in a root directory from an mu-plugin:

```
require_once BEDROCK_ROOT_DIR . '/bootstrap/app.php';
```

Another byproduct is that `application.php` and other "core" configuration modules will most likely not need to be changed and custom user configuration can simply be another drop-in module.

Any required temporary variables are also unset now where possible to prevent unnecessary leakage into the global scope.

**A note about the usage of `glob()`**

This change leverages the [glob()](http://php.net/manual/en/function.glob.php) function to read in the files from the new `config/modules` directory.  Some might be concerned about this as an unnecessary performance cost on every request, but from my research it's an amount of insignificant time.

There is an interesting report on it [here](http://jameshalsall.co.uk/posts/glob-performance-for-finding-files-in-php) although it is roughly 3 years old (read: way before PHP 7).  Even at this time, with a set of thousands of files of mixed types, globbing all php files took roughly a _half of a hundredth of a second_.  You can actually clone the repo and run their benchmarks yourself quite easily.

Anyways, this could enhanced for performance with caching if needed, it just seems unnecessary and would come at the cost of increased complexity.

### DIFF
--- a/config/application.php
+++ b/config/application.php
@@ -1,10 +1,15 @@
 <?php
 
 /** @var string Directory containing all of the site's files */
-$root_dir = dirname(__DIR__);
+define('BEDROCK_ROOT_DIR', dirname(__DIR__));
 
 /** @var string Document Root */
-$webroot_dir = $root_dir . '/web';
+define('BEDROCK_WEB_ROOT_DIR', BEDROCK_ROOT_DIR . '/web');
+
+/** @var string Absolute path to the core WordPress directory */
+if (!defined('ABSPATH')) {
+    define('ABSPATH', BEDROCK_WEB_ROOT_DIR . '/wp/');
+}
 
 /**
  * Expose global env() function from oscarotero/env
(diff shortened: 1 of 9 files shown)


swalkinshaw:
I like the spirit behind this PR since I'm a fan of organization, but I don't think the config is the best place for it.

There's definitely "sections" to the config so I can see why you'd want to separate them out. But Bedrock already has 1 level of indirection with `wp-config.php` and I don't really want to add another one.

I'm all in favour of users just including/requiring additional config files as they see fit though.

A lot of the concepts behind Bedrock come from Rails. In Rails for example, there's `config/application.rb` and then the same environments files that Bedrock has. Rails then has "initializers" in `config/initializers` which act like the modules you introduced here. mu-plugins are kind of analogous to them.

I've actually seen another WP project like Bedrock that has `config/initializers`. While I probably wouldn't add them to Bedrock itself, I think this pattern of "module" files fits better in that concept than for the config file.

aaemnnosttv:
Thanks for considering my PR Scott.  It sounds like I don't have your vote just yet but you didn't close it in my face (appreciated).. allow me to try to make my case.

> Bedrock already has 1 level of indirection with wp-config.php and I don't really want to add another one.

Why is that? If there was no benefit then I would totally understand, but it's not organization just for the sake of organization;  it's an effort to make Bedrock extensible.  Couldn't you make the same argument for not moving templates into their own folder in Sage 9?  I love that change, even at the cost of the indirection (which is much more substantial IMO seeing as it has to change an option value to work).  To your point, the remaining parts to `application.php` could also be moved to modules, or to `wp-config.php`, removing the need for `application.php` altogether, and thus keeping the level of indirection the same; I think the way it is in the PR is better though.

I'm all for taking inspiration from mature, widely-used frameworks like Rails, but I think it would be a mistake to try to _mirror_ them because there is a big difference in context.  Laravel (also heavily Rails-inspired) [separates the configuration into pieces](https://github.com/laravel/laravel/tree/master/config), although differently.

There may be a better name for the `modules` folder, it's not the first name I thought of either.  If there is another convention that is ubiquitous or more intuitive, I would be all for it; `initializers` in this context does not sound that intuitive to me though because they don't initialize anything.

Regarding the concept of initializers, there's a big difference between one of these module files and an mu-plugin (ie: all of WordPress core loading).  Timing is key, and capability is much different.

Apart from that, (and this could be a separate PR of course) but the constants I introduced would also be valuable, I don't see any reason not to use those.  Once you no longer have a need for the temporary variables, there is also no need to keep everything in one file like that either.

I think it's also worth considering the upgrade path for the user, as Bedrock is a project that has been evolving over time as well (which is part of what makes it great!).  It's much easier IMO to replace a few files than to diff a bigger one and upgrade via surgery.  By encouraging users to add customizations to different files, this would essentially eliminate the need to do any manual editing of "core" Bedrock files/config.

I hope that all makes sense.  Thanks again for reading through all that, and your time in corresponding.

swalkinshaw:
Well I wouldn't compare this to what Sage 9 is doing. Different projects, contexts, etc. Also I don't really like what's needed to make that template folder work either :)

I agree this does help with customization and extensibility but it doesn't solve it either. I also don't really like the number prefix to enforce include order (but I realize there's no other solution for that).

Anyway, if this is merged it won't be right away. Need to sit on it for a bit either way.

aaemnnosttv:
Thanks Scott.  I'm curious what you mean about "solving" extensibility?  I don't mean to sound facetious, but how would it look differently if it were solved?  Is it possible to solve?  I think being able to drop in a file and "it just works" is a pretty good experience all around :D

If you and the team decide it's not the right fit for Bedrock, I respect that; just trying to understand all perspectives.

swalkinshaw:
I doubt it can be solved completely. I just meant that these configs would definitely make upgrades easier, but conflicts could still easily happen.

swalkinshaw:
@aaemnnosttv I took another look at this recently and we're going to pass on it. Thanks for proposing it. I just don't think there's enough benefit at this point to warrant the change.

PR #271, opened 2016-08-08
Add environment vars for database collation, charset

michaeldyrynda:
Update application config to use the new environment configuration variables.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2,6 +2,8 @@ DB_NAME=database_name
 DB_USER=database_user
 DB_PASSWORD=database_password
 DB_HOST=database_host
+DB_CHARSET=utf8mb4
+DB_COLLATE=
 
 WP_ENV=development
 WP_HOME=http://example.com
--- a/config/application.php
+++ b/config/application.php
@@ -52,8 +52,8 @@ define('DB_NAME', env('DB_NAME'));
 define('DB_USER', env('DB_USER'));
 define('DB_PASSWORD', env('DB_PASSWORD'));
 define('DB_HOST', env('DB_HOST') ?: 'localhost');
-define('DB_CHARSET', 'utf8mb4');
-define('DB_COLLATE', '');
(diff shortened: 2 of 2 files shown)


swalkinshaw:
What's the use case for this? The idea behind config settings with env vars are:

1. They are "secret" and shouldn't be committed
2. They vary between environment

These definitely aren't secrets and they really *shouldn't* differ per environment. It could easily lead to issues if they did.

michaeldyrynda:
I just felt it odd that everything else was configurable via env except these parameters, but I can see your point of view as well.

It was only because I was using [Herbert](http://getherbert.com/) and it was having issues with utf8mb4 collation that I went looking for a solution.

I suppose it's not something most would have to change, would some clearer documentation on this be more suitable?

swalkinshaw:
Documentation around `utf8mb4` specifically? `utf8mb4` should the default nowadays and even WP [upgrades](https://make.wordpress.org/core/2015/04/02/the-utf8mb4-upgrade/) when its available.

eimanavicius:
Why DB_PREFIX is not in .env.example file? But have environment var?

Maybe we should have these new vars with defaults?

QWp6t:
@eimanavicius https://github.com/roots/bedrock/pull/261

QWp6t:
I'm with swalk on this, but if there's demand for it, then it should follow the pattern in #261 for optional variables, i.e., it exists in the .env.example file, but it's commented out, and there is a default value for it defined in application.php. But again, that's only if there's a demand. These are variables that really shouldn't be changed between environments.

swalkinshaw:
Yeah I'm going to close this in its current state for the reasons mentioned above.

michaeldyrynda:
Thanks, folks. With regards to docs, I was referring to the additional variables that could be changed in `config/application.php`, but looking [here](https://roots.io/bedrock/docs/configuration-files/) I think there's probably sufficient info there.

PR #273, opened 2016-08-12
Use site transients for saving mu-plugins

rasmusbe:
Since the mu-plugins list should be considered a cache we should use transients for saving the plugin list.

### DIFF
--- a/web/app/mu-plugins/bedrock-autoloader.php
+++ b/web/app/mu-plugins/bedrock-autoloader.php
@@ -112,7 +112,7 @@ class Autoloader
      */
     private function checkCache()
     {
-        $cache = get_site_option('bedrock_autoloader');
+        $cache = get_site_transient('bedrock_autoloader');
 
         if ($cache === false) {
             $this->updateCache();
@@ -138,7 +138,7 @@ class Autoloader
         self::$activated    = ($rebuild) ? $plugins : array_diff_key($plugins, self::$cache['plugins']);
         self::$cache        = array('plugins' => $plugins, 'count' => $this->countPlugins());
 
-        update_site_option('bedrock_autoloader', self::$cache);
+        set_site_transient('bedrock_autoloader', self::$cache);
     }
 
     /**


swalkinshaw:
@Foxaii

Foxaii:
Sorry to have taken so long to respond to this issue.

I made the conscious decision to use a site option when writing the original plugin. I did so because the list of mu-plugins isn't invalidated over time: it's invalidated when plugins are added or removed. Standard plugins are also saved as options, not transients, although I appreciate there is a meaningful difference between the two because standard plugins are an option in the WordPress backend.

There's little or no difference between options and transients in practice, although I think that if we switch (which I'm not inclined to do) we need to add a comment to make it clear it's an autoloaded transient, without an expiry date.

Do you have anything to add to support the switch?

QWp6t:
With persistent cache plugins, transients can be handled by, for example, Redis. Theoretically this means that the data can be retrieved faster, but I remain skeptical of the actual impact of what ostensibly seems to be a micro-optimization given the small size of the data that's stored by the autoloader. It also makes the presence of the cache less predictable, which means it might need to be regenerated more frequently, i.e., transient data can be purged at any time. Having it as an option makes its presence marginally more predictable. But ultimately, I don't think either of these are strong enough arguments for either side, and I would just as soon defer my position on this issue to @foxaii for his reasoning.

tl;dr-

 ¯\\\_(ツ)_/¯

rasmusbe:
Thank you for your reply @Foxaii 

Like @QWp6t said, we do use Redis as object cache. And we are deploying by using Capistrano. We do have a Capistrano task to flush the cache when deploying (and for example adding new mu-plugins) and right now the mu-plugin cache is not invalidated, therefore generating some trouble for us.

With this change the mu-plugin cache is invalidated correctly when other cache values is flushed.

Foxaii:
The list of mu-plugins is updated whenever a path is missing (i.e. when a plugin is removed or renamed) or when a new folder is found (i.e. when a plugin is added), so you really shouldn't need to be flushing anything manually, whether it's an option or transient.

Do you mind elaborating on the 'trouble' you are seeing as it could be a bug.

rasmusbe:
I'm sorry for a slow reply.

If we add a new mu-plugin it won't be loading after deploy since it's not in the site option and it won't be loaded until someone visits the wp-admin plugin page.
If the plugin uses transients instead and we flush the object cache when deploying it will be loaded.

rasmusbe:
@Foxaii

Foxaii:
Sorry for the delay in getting to this.

I understand that manually flushing the transient is solving the problem for you @rasmusbe, but the underlying issue is that the plugin is not correctly invalidating its cache.

We can either fix the plugin's cache invalidation, which fixes the issue for everyone (whether it uses an option or transien), or switch to a transient, which fixes it only for people who flush caches after a deploy.

That's why I asked for more data from you, as the mu-plugin cache should be invalidating itself if a plugin isn't found, or the number of folders have changed. If we can figure which of these events isn't being detected correctly on your project, we can fix the underlying issue.

I hope that makes sense.

swalkinshaw:
Closing this as @Foxaii's last comment makes it clear that this doesn't solve the actual problem.

PR #276, opened 2016-09-27
Fix URL issues on multisite: Part 2

ryanplasma:
This PR picks up on the work by @felixarntz in roots/bedrock#251 by fixing the PHPCS errors

My (non-extensive) testing so far shows that this fixes the multi-site link issues.  Open to discussion, but at least now it passes travis tests.


eimanavicius:
wp dir name should be changeable from constants or similar

PR #277, opened 2016-11-13
Production environment by default

kalenjohnson:
In my enormous amount of spare time, I've been working on a side project where I'm working on integrating best practices throughout a WP project.

It seems like a more sensible default for a site would be to default to production, rather than development. Development has all errors being displayed, amongst some other things. It's rather pedantic, but if a user forgets to set this for some reason, the safer option would be production rather than development

### DIFF
--- a/config/application.php
+++ b/config/application.php
@@ -22,9 +22,9 @@ if (file_exists($root_dir . '/.env')) {
 
 /**
  * Set up our global environment constant and load its config first
- * Default: development
+ * Default: production
  */
-define('WP_ENV', env('WP_ENV') ?: 'development');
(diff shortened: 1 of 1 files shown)


kalenjohnson:
And to be clear, Laravel triggered this idea for me: https://github.com/laravel/laravel/blob/master/config/app.php#L28

swalkinshaw:
I think the idea behind this was that a lot of people were trying out Bedrock/Trellis on development first, and potentially only using them for that purpose so this default reduced a potential extra config setting people had to remember to set.

Probably a good thing to change it though 👍

PR #283, opened 2016-12-06
Move PHPCS settings to xml config.

schrapel:


swalkinshaw:
Makes sense. Thanks 👍

PR #290, opened 2017-01-14
Fix composer version constraint warnings

garak:
Old composer has following issues:
- if your run `composer outdated`, you can find that `composer/installers` is at version `v1.2.0`, while installed is `v1.0.25`
- if you run `composer validate composer.json`, you can find that "exact version constraints should be avoided"

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,11 +34,11 @@
   ],
   "require": {
     "php": ">=5.6",
-    "composer/installers": "~1.0.12",
+    "composer/installers": "^1.2",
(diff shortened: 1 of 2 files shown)


swalkinshaw:
We usually prefer exact versions but if composer itself is going to recommend against it then maybe we should change it.

Regardless of that, this PR would need updates `composer.lock` as well since the `composer/installers` version was changed.

swalkinshaw:
Took care of composer/installers in https://github.com/roots/bedrock/pull/306.

Still not sure about WP constraints.

PR #292, opened 2017-01-17
xdebug is missing for nightly

sergeyz:


swalkinshaw:
There's like 4 changes in here...

1. I'm not sure we want to test on HHVM.
2. Nightly should still be allowed to fail
3. Allow fast failures seemed like a good thing that was removed?
4. Xdebug fix: good 👍

sergeyz:
@swalkinshaw fixed

swalkinshaw:
Perfect, thanks 👍

PR #297, opened 2017-02-09
Relax version constraint for WordPress core

dwalkr:
Users should be able to apply minor and patch updates to core with a simple `composer update`


QWp6t:
WordPress doesn't follow semver. The first two version sequences designate the major release. The third sequence is minor. (They don't have patch releases.) If we are going to relax the version constraint, then I think we should leave off the third sequence entirely and just go with...
```
  "johnpbloch/wordpress": "4.7",
```

swalkinshaw:
Not really to relax these version constraints as @QWp6t mentioned

PR #312, opened 2017-04-24
Ignore CHANGELOG.md on composer create command

ankurk91:
Composer can ignore this file by reading .gitattributes


swalkinshaw:
This was purposely left out (of gitattributes) along with the `README`. We think those two files are valuable and should always exist.

ankurk91:
Ok, no problem.

PR #313, opened 2017-04-24
Auto create .env on composer create

ankurk91:
We know that everyone going to copy ```.env.example``` and create a ```.env``` so why not auto create this file.
This script will run only when someone​ use ```composer create``` command.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,6 +52,9 @@
     "wordpress-install-dir": "web/wp"
   },
   "scripts": {
+    "post-root-package-install": [
+      "php -r \"copy('.env.example', '.env');\""
+    ],
     "test": [
       "vendor/bin/phpcs"
     ]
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -52,6 +52,9 @@
     "wordpress-install-dir": "web/wp"
   },
   "scripts": {
+    "post-root-package-install": [
+            "php -r \"file_exists('.env') || copy('.env.example', '.env');\""
```
swalkinshaw: Is the check for an existing file even necessary? Since this only happens on `create-project`, it's always a new project anyway.

This formatting is off as well, you need to match the indentation with the existing file.

ankurk91: Updated, please check, feel free to modify in my repo, see [how](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/)

swalkinshaw:
Works, thanks!

PR #315, opened 2017-05-03
Bump WordPress to 4.7.4.1

austinpray:


swalkinshaw:
Note: this is **not** an actual WordPress version. It's a version from the `johnpbloch/wordpress` composer package which fixed how its releases get done.

More details here: https://github.com/johnpbloch/wordpress/issues/32

PR #316, opened 2017-05-16
Remove /wp/ from no wp-admin/wp-includes URIs from site_url()

antoscarface:
Anyone have understood well the difference between home_url() and site_url() yet.

So, the results is that with bedrock (in this case) often I have to see URIs on frontend like that: `mysite.com/wp/page` instead of `mysite.com/page`

Many plugins still use site_url() instead of home_url() for the URLs and you will see may frontend URLs with `/wp`.

I thought: the `/wp` is necessary only because of the folders `wp-includes` and `wp-admin` are inside the `wp` folder. So, my solution is to add a mu-plugin that remove the `/wp` part from all URLs passed from `site_url` that don't have `wp-admin` or `wp-includes` part and remove the `/wp` part.

It seems work well. Do you have some contraindication about this that I'm missing?

I hope the explanation is clear for you :)

### DIFF
--- a/web/app/mu-plugins/site-url-fix.php
+++ b/web/app/mu-plugins/site-url-fix.php
@@ -1,0 +1,26 @@
+<?php
+/*
+Plugin Name:  Site URL fix
+Plugin URI:   https://roots.io/bedrock/
+Description:  Remove /wp/ if the URL doesn't contain /wp-includes/ or /wp-admin/
+Version:      1.0.0
+Author:       Roots
+Author URI:   https://roots.io/
+License:      MIT License
+*/
+
+namespace Roots\Bedrock;
+
+if (!is_blog_installed()) {
+    return;
+}
+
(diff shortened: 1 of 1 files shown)


retlehs:
thanks for the PR! rather than encouraging wordpress plugins to continue to use the incorrect function, why not reach out to the plugin authors and have them fix it?

there's tons of plugins that do this, hardcode wp-content, and other things that break if you use bedrock (or wordpress in a subdirectory, or rename the wp-content directory). at the end of the day these are issues with the plugins. if a plugin doesn't work in bedrock, it's a good indication that there's something wrong with the plugin.

antoscarface:
I already do this :) I already reach authors, but no answers. For now, I had to put this fix on my website, otherwise I will have problems with almost every plugin. 

So, maybe this fix could be a suggested extension for bedrock?

PR #320, opened 2017-06-23
Added env constants for custom application settings

ruscon:


QWp6t:
i don't really see the use case where these values would change between environments tbh

ruscon:
> between environments tbh
May be is not between environments. 
But they hard coded :)
In some cases they needed to be edited.

TangRufus:
+1

Use case:
A throw away server for clients to play with their plugins.

Think you are managing client's sever. Instead of changing `composer.json` every time clients want to play with their new found toys, we can create another less restrictive staging for them. Allowing clients to install plugins by themselves.

austinpray:
Wow I apologize for leaving this PR to rot. Sorry to say that this is definitely a non-starter for me.

1. Nearly everything should be statically defined in your config/application.php. The only things that should change between environments are credentials and things like error reporting levels. Making these togglable via env vars goes against this.
2. If you want a special staging environment create an environment specific config file called config/staging_special.php, set wp_env to staging_special and then override things as you please.

vpillinger:
``
If you want a special staging environment create an environment specific config file called config/staging_special.php, set wp_env to staging_special and then override things as you please.
``

Doing this will cause PHP to throw notices because constants are being redefined.
image

austinpray:
Oh yep sorry, good point. My mind was not in in Bedrock/WordPress land. I agree that the current configuration behavior of:

image

Is ridiculous compared to say laravel, django, or rails

```php
config(['app.timezone' => 'America/Chicago']);

$value = config('app.timezone');
```

```python
from django.conf import settings

settings.DEBUG = True  
```

```ruby
# config/environments/production.rb
config.time_zone = 'Central Time (US & Canada)'
```

I'm gonna be working on this this weekend

Issue #323, opened 2017-07-17
Using oscarotero/env in situations where putenv()/getenv() is disabled


petrpacas:
- [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md)
- [ ] This is a feature request
- [ ] This is a bug report
- [x] This request isn't a duplicate of an [existing issue](https://github.com/roots/bedrock/issues)
- [x] I've read the [docs](https://roots.io/bedrock/docs) and followed them (if applicable)

---

**What is the current behavior?**

I'm trying to use Bedrock in a shared webhost and as it turns out, they have `putenv()` (and consequently `getenv()`) disabled.

`$_ENV[]` works just fine.

Unfortunately https://github.com/oscarotero/env/blob/master/src/Env.php#L37 uses `getenv()` in it's function.

**What is the expected or desired behavior?**

https://github.com/roots/bedrock/pull/229 tried to suggest an alternative solution for the `true/false` string dillema. Couldn't we get back to it to prevent an issue that people might eventually run into?

_(I made https://github.com/oscarotero/env/issues/3 just in case...)_

petrpacas:
(I made https://github.com/oscarotero/env/issues/3 just in case...)

petrpacas:
Right, so the library's creator will release a new version that will have an option to get the values from `$_ENV[]` instead of `env()`.

```
Env::init();
Env::$options = Env::USE_ENV_ARRAY;
```

PR #324, opened 2017-07-17
Bump oscarotero/env

petrpacas:
Closes https://github.com/roots/bedrock/issues/323

Bumping [oscarotero/env](https://github.com/oscarotero/env) to a newly released version would help with possible cases where the `env()` function wouldn't work in cases where server has disabled `putenv()` + `getenv()` commands.


swalkinshaw:
Thanks 👍

PR #338, opened 2017-12-10
Call composer package binaries without path

LeoColomb:
Exactly the same as https://github.com/roots/sage/pull/1846

See composer note at https://getcomposer.org/doc/articles/scripts.md#writing-custom-commands


swalkinshaw:
Thanks!

PR #349, opened 2018-02-11
Add php 7.2 to test matrix

mvasin:


swalkinshaw:
Thanks!

mvasin:
I have also noticed that there are in fact no tests. I reasonable test suite would include `composer install`, bringing a WordPress installation to life and running at least a few end-to-end tests, to make sure WordPress is operational. The test suite would also document what should be done to start using Bedrock. Code sniffing is nothing.

swalkinshaw:
Yeah, correct. The current Travis set up is basically the minimum we could do (beyond literally nothing) and may only catch a tiny percentage of issues. I personally looked into doing exactly what you described a few times but honestly forget why it was never completed (beyond laziness).

Feel free to contribute that test suite if you want 👍

PR #363, opened 2018-07-09
Manage languages through composer and https://wp-languages.github.io

kgoedecke:
I would like to suggest that we integrate https://wp-languages.github.io into bedrock. We've had a lot of issues managing languages. I think this should be quite helpful, since during a deployment the languages simply get added automatically and you don't depend on WordPress automatically downloading them. 

It would be great if we could discuss the pros/cons of adding this PR.

Cheers

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@
 web/app/plugins/*
 !web/app/plugins/.gitkeep
 web/app/mu-plugins/*/
+web/app/languages/*
+!web/app/languages/.gitkeep
 web/app/upgrade
 web/app/uploads/*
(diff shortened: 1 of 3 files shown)


austinpray:
I appreciate this PR! In its current form I'm not sure it will work though. We slimmed down our setup script to make cloning and `composer create` methods of creating a bedrock project first class citizens. The default language selection would _have_ to be interactive. Even `core-en_gb` as a default is contentious.

1. You should definitely turn this PR into a [guide](https://roots.io/guides/)
2. Maybe we should make a package that metaprograms these changes after interactively prompting a user for a language?

PR #366, opened 2018-07-10
Add WP PHPUnit

### LINKED ISSUE #365: Feature Request: Add first-class support for PHPUnit
## Submit a feature request or bug report

- [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md)
- [x] This is a feature request
- [ ] This is a bug report
- [x] This request isn't a duplicate of an [existing issue](https://github.com/roots/bedrock/issues)
- [x] I've read the [docs](https://roots.io/bedrock/docs) and followed them (if applicable)

---

## Feature Request

Installing, configuring, and running the WordPress core PHPUnit library has always been a pain. This creates an unnecessary barrier for WordPress developers to embrace testing, which is undoubtedly a core component of writing modern web applications.

I've been working on the best way to make the WordPress core PHPUnit library installable via Composer for a while now, and I'm happy to say that it is now just as easy to install as a vanilla Composer package.

I would like to propose the addition of [WP PHPUnit](https://github.com/wp-phpunit) to Bedrock's development dependencies, along with the necessary boilerplate for everything else PHPUnit needs to run - including CI configuration. Once Composer dependencies have been installed, and the database created, PHPUnit can be run with a single command out-of-the-box.

I would liken the end goal to be similar to what is part of a fresh installation of Laravel. Structure, dependencies, and a working example test.

Unit tests are not just for plugins/themes. Applications like those built on Bedrock will likely have code which is specific to the application, or not structured as a plugin or theme which should still be tested. Not only that, but testing that your code works as expected in conjunction with all the other dependencies that your application requires is arguably more important than testing that a specific component of it works in isolation.

WP PHPUnit is extremely minimal and is very easy to use compared to the current workflow. It's also fully automated in its maintenance, and completely open-source.

You can see a full working example here: https://github.com/wp-phpunit/example-project 
Although this isn't based on Bedrock, the setup is even simpler _with_ Bedrock 😄

If added, WP PHPUnit could be easily removed if undesired without requiring any kind of complicated uninstallation procedure, similar to other components of Bedrock. I think I've made the case that its value as an easily opt-out-able component outweighs the cost one would incur if it were opt-in instead.

Is this something that the team would be receptive to considering a PR for?

Related #106

aaemnnosttv:
This is a first draft of a working implementation of what it would look like to integrate WP PHPUnit into Bedrock.

Documentation is not included yet and still needs to be added, but I wanted to get a first round of feedback before writing it.

There are a few things that are somewhat of stylistic preferences, like the exceptions made to the PHPCS rules for tests and the formatting of the example test. This is somewhat bending the rules of the existing ruleset, but I believe it's a common test-specific style.

Apart from that, very little has really changed, as almost everything here are additions.

I look forward to hearing anyone's feedback on the proposed changes.

Resolves #365


.travis.yml
```diff
@@ -18,12 +18,23 @@ cache:
...
 before_install:
   - phpenv config-rm xdebug.ini || true
   - composer self-update
 
 install:
   - composer install -o --prefer-dist --no-interaction
 
+before_script:
+  - mysql -u root -e "GRANT ALL PRIVILEGES ON ${DB_NAME}.* TO ${DB_USER} IDENTIFIED BY '${DB_PASSWORD}';"
+  - mysql -u root -e "CREATE DATABASE ${DB_NAME};"
```
austinpray: Thinking maybe travis db testing integration should be a blog post or something separate. There are many ways to skin this cat (docker, external dbs, etc) so best to leave up to the user's discretion. We shouldn't even need a db for unit tests, that is for e2e tests.

composer.json
```diff
@@ -55,8 +60,12 @@
     "post-root-package-install": [
       "php -r \"copy('.env.example', '.env');\""
     ],
-    "test": [
-      "phpcs"
-    ]
+    "test": "phpunit",
+    "lint": "phpcs"
```
austinpray: this is a good change

config/application.php
```diff
@@ -72,7 +72,6 @@
  * Custom Settings
  */
 define('AUTOMATIC_UPDATER_DISABLED', true);
-define('DISABLE_WP_CRON', env('DISABLE_WP_CRON') ?: false);
```
austinpray: what is going on here?

composer.json
```diff
@@ -41,7 +44,9 @@
     "roots/wp-password-bcrypt": "1.0.0"
   },
   "require-dev": {
-    "squizlabs/php_codesniffer": "^3.0.2"
+    "phpunit/phpunit": "^5.0",
+    "squizlabs/php_codesniffer": "^3.0.2",
+    "wp-phpunit/wp-phpunit": "4.9.7"
```
austinpray: Honestly don't really want to be married to wp-phpunit. We should leave this open ended and just give people raw phpunit and let people choose whatever approach they want. eg https://github.com/10up/wp_mock type approach or full on unit + e2e tests and such.

@johnpbloch what do you think?

austinpray:
Also wow sorry meant to include this in initial review but: thank you so much for putting this together! This is a great start :+1:.

johnpbloch:
I haven't taken a look at the code, just jumping in to follow up to @austinpray's comment. I think insofar as the PR improves the compatibility between WP PHPUnit and Bedrock, this is a great initiative. I do tend to agree with the idea of keeping Bedrock testing framework-agnostic. Maybe it would be worth putting some of the more popular testing frameworks (WP Mock, WP PHPUnit, BrainMonkey, etc.) in the suggest block of `composer.json` with descriptions of their intended use? Just a thought.

swalkinshaw:
I mostly agree with @austinpray. I kind of forgot that the WP test unit library is mostly designed for testing WP core itself (for core developers). It's pretty heavy-handed for normal application/project testing.

So having the boilerplate required to make that easier with Bedrock is still useful 👍

config/application.php
aaemnnosttv: This is not completely removed, but the define is moved into the environment configuration files. This is because this constant cannot be defined in the testing environment, as it will define it itself without checking if it has already been defined or not. 

The alternative to moving it to the configuration files (and not to `testing.php`) is to wrap this with a conditional to only define it if `WP_ENV !== 'testing'` which is a little less clean IMO.

composer.json
aaemnnosttv: That's a fair statement, and I'm probably biased as the package creator, but wp-phpunit is more practical for the average user as it will be what most people are familiar with IMO. I don't think it would be a problem for anyone who is familiar with other testing frameworks to replace and use whatever they want.

I think reducing the setup to bare PHPUnit would be less valuable for more people, and wouldn't really make a difference to those who are want to use something else. It's just a boilerplate. No one is married to anything.

Of course bare PHPUnit is better than nothing, I think we agree on that.

composer.json
```diff
@@ -24,7 +24,10 @@
     "forum": "https://discourse.roots.io/category/bedrock"
   },
   "config": {
-    "preferred-install": "dist"
+    "preferred-install": "dist",
+    "platform": {
+      "php": "5.6.32"
```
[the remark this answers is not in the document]
aaemnnosttv: Agreed. This has to do with PHPUnit more than anything. The reason this is here is so that all dependencies installed will be compatible with PHP 5.6 when it runs on Travis. This is because, even though PHPUnit 5.x is required, because my local version of PHP is 7.2, some of PHPUnit's dependencies will be installed with versions which are not compatible with 5.6. 

The alternatives to doing this are:
- Requiring in the offending dependencies and locking them to a version which is compatible with PHP 5.6. Not a good option as those dependencies could always change and this would be a pain to manage.
- Running `composer update` on Travis instead of `composer install`. I don't like this as much either as its less predictable
- Always run `composer update` with PHP 5.6. 

This is a very useful configuration would be something that could be added to the docs, as most users probably aren't using Bedrock in a multi-version environment. Users can (and probably should be) set this to their own value based on their target environment.

With PHP 5.6 and 7.0 reaching end of life at the end of the year perhaps this won't be necessary anymore.

aaemnnosttv:
Thank you everyone for your feedback! It seems like everyone is on board with the idea that adding some boilerplate for testing is a step in the right direction but maybe it shouldn't be opinionated in choosing a framework to include boilerplate for.

> PHPUnit + wp-phpunit is the best way to test your wordpress installation.

I don't particularly agree with the idea that including a library is making such a statement, nor would I make that claim myself.

The library is really just the WordPress core library that supports running PHPUnit in a WordPress environment. It's just been made easily installable via Composer. I would say that this library is the one that most people are going to be familiar with in a WordPress context (WP CLI scaffolds it, and most articles/resources on testing in WordPress are based on it).

Of course others are available, but I think if Bedrock were to include a more complete testing boilerplate (as proposed) that it offers more value than if it was more barebones. By no means does it lock anyone in to using it or impose anything that isn't easily changeable which I think is the important aspect of the principal of remaining framework agnostic.

With that said, I fully respect the teams decision to keep Bedrock framework agnostic if that's the consensus. In my opinion, including the "official" core library for testing is still relatively framework agnostic because it's part of WordPress core.

Either way, I'm still interested in helping with this initiative in whatever way I can.

99linesofcode:
This has been stale for quite some time and there has been much discussion on the subject prior to this PR. What can I do to help this along as I would love to be able to write unit tests for our suite of plugins.

chrillep:
@aaemnnosttv could make an example project of this instead ?
let's say wp-phpunit/example-bedrock-project ?

Issue #299, opened 2017-02-22
[Feature Request] Move Autoloader to its own repository


vinkla:
It would be nice if we could use the autoloader outside of Bedrock. Would it be possible to move the class to its own Composer package or its own plugin?

swalkinshaw:
Agreed but unfortunately I don't think it's possible :(

The autoloader is needed so that mu-plugins via Composer work, so it can't be a package itself. And it could be a normal plugin, but we only use Composer to manage plugins in Bedrock.

vinkla:
Oh, okay. Too bad :(

aaemnnosttv:
I've been wondering the same for a while now and think this is worth revisiting.

@swalkinshaw it can't be entirely implemented as a package without using a custom loader, etc, but it should be entirely possible to install `Autoloader` as a regular package that lives in the vendor directory. Then the mu-plugin in Bedrock could be drastically simplified like so:

```php
/** file header... */

namespace Roots\Bedrock;

if (is_blog_installed() && class_exists(Autoloader::class)) {
    new Autoloader();
}
```

So, I think even though there would still be some very minimal "wiring" necessary to make it work, it would still be valuable as a package. It's also a substantial chunk of the project which isn't really intended to be boilerplate.

Log1x:
I think this is a good idea and I'd be willing to rewrite the autoloader (or at the very least, clean it up).

QWp6t:
we could write a composer plugin that copies the autoloader to mu-plugins folder

aaemnnosttv:
@QWp6t that sounds a bit unnecessary considering the simplicity of the integration. Then you would be looking at 3 packages instead of one:

* Autoloader package
* Bedrock Autoloader installer Composer plugin
* Bedrock Autoloader package that uses both

I think a Composer plugin would be more justified if the mu-plugin required some dynamic component that the plugin would generate.

austinpray:
Started work on this: https://github.com/roots/bedrock-autoloader

I made all the proper sacrifices to the git gods such that the autoloader's commit history made it to the new repo :+1:

swalkinshaw:
I agree this is a worthwhile improvement 👍

austinpray:
[Making issues here](https://github.com/roots/bedrock-autoloader/issues) as they come to mind. Feel free to grab some of them off the stack! @aaemnnosttv @vinkla and other friends: feel free to add more issues to the pile if I am missing anything

vinkla:
[...]

PR #369, opened 2018-07-31
Show admin notice when indexing is discouraged

### LINKED ISSUE #368: Undo #59 and never prevent indexing or move it to a separate config
## Submit a feature request or bug report

- [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/guidelines/blob/master/CONTRIBUTING.md)
- [x] This is a feature request
- [ ] This is a bug report
- [x] This request isn't a duplicate of an [existing issue](https://github.com/roots/bedrock/issues)
- [x] I've read the [docs](https://roots.io/bedrock/docs) and followed them (if applicable)

Replace any `X` with your information.

---

**What is the current behavior?**

When the site is in not in production mode indexing is prevented by a wordpress hook. This also prevents indexing in say case where environment is accidentally set to "productoin". Which is undesirable because a simple mistake or typo can silently wreck an seo ranking.

**What is the expected or desired behavior?**

The site should never prevent indexing unless explicitly told to and with a warning displayed.

Some options include 1) Moving it to a separate config and displaying a warning in wp-admin when it is not in allowing indexing, or 2) removing it entirely. 

---

## Feature Request

**Please provide use cases for changing the current behavior:**

1) This is a potential source of serious SEO trouble. Inexperienced developers, unattentive server admins, or just human error make it very likely that a site's SEO ranking gets wiped. There is no visible warning or display that the site is not indexing information, and in the kind of teams that make these configuration mistakes, it is unlikely to be found until it is too late.

2) It is better to disallow indexing and protect from access by bots by using basic auth. Unsavory bots can still find your site and ignore the meta tag with the current implementation. Basic auth is also much more visible and will be corrected quickly in production.

**Other relevant information:**

QWp6t:
I haven't tested this yet. I'll leave that to someone else (@vpillinger?) because I'm lazy.

Closes #368

### DIFF
--- a/web/app/mu-plugins/disallow-indexing.php
+++ b/web/app/mu-plugins/disallow-indexing.php
@@ -17,6 +17,10 @@ if (!defined('WP_ENV') || WP_ENV === 'production') {
 add_action('pre_option_blog_public', '__return_zero');
 
 add_action('admin_notices', function () {
-    $message = __('Search engine indexing has been discouraged because the current environment is', 'roots');
-    echo "<div class='notice notice-warning'><p><strong>Bedrock:</strong> {$message} <code>".WP_ENV."</code>.</p></div>";
+    $message = sprintf(
+        __('%1$s Search engine indexing has been discouraged because the current environment is %2$s.', 'roots'),
+        '<strong>Bedrock:</strong>',
+        '<code>'.WP_ENV.'</code>'
+    );
+    echo "<div class='notice notice-warning'><p>{$message}</p></div>";
 });
--- a/web/app/mu-plugins/disallow-indexing.php
+++ b/web/app/mu-plugins/disallow-indexing.php
@@ -6,6 +6,7 @@ Description:  Disallow indexing of your site on non-production environments.
 Version:      2.0.0
 Author:       Roots
 Author URI:   https://roots.io/
+Text Domain:  roots
 License:      MIT License
 */
 
@@ -18,6 +19,6 @@ if (!is_admin()) {
 } else {
     add_action('admin_notices', function () {
         $message = __('Search engine indexing is currently discouraged.', 'roots');
-        echo sprintf('<div class="error"><p>%s</p></div>', $message);
+        echo sprintf('<div class="notice notice-warning"><p>%s</p></div>', $message);
     });
 }
--- a/web/app/mu-plugins/disallow-indexing.php
+++ b/web/app/mu-plugins/disallow-indexing.php
@@ -18,5 +18,5 @@ add_action('pre_option_blog_public', '__return_zero');
 
 add_action('admin_notices', function () {
     $message = __('Search engine indexing has been discouraged because the current environment is', 'roots');
-    echo "<div class='notice notice-warning'><p><strong>Bedrock:</strong> {$message} <code>".env('WP_ENV')."</code>.</p></div>";
+    echo "<div class='notice notice-warning'><p><strong>Bedrock:</strong> {$message} <code>".WP_ENV."</code>.</p></div>";
 });
--- a/web/app/mu-plugins/disallow-indexing.php
+++ b/web/app/mu-plugins/disallow-indexing.php
@@ -3,12 +3,21 @@
 Plugin Name:  Disallow Indexing
 Plugin URI:   https://roots.io/bedrock/
 Description:  Disallow indexing of your site on non-production environments.
-Version:      1.0.0
+Version:      2.0.0
 Author:       Roots
 Author URI:   https://roots.io/
 License:      MIT License
 */
 
-if (defined('WP_ENV') && WP_ENV !== 'production' && !is_admin()) {
+if (!defined('WP_ENV') || WP_ENV === 'production') {
+    return;
+}
+
+if (!is_admin()) {
     add_action('pre_option_blog_public', '__return_zero');
+} else {
+    add_action('admin_notices', function () {
+        $message = __('Search engine indexing is currently discouraged.', 'roots');
+        echo sprintf('<div class="error"><p>%s</p></div>', $message);
+    });
 }
--- a/web/app/mu-plugins/disallow-indexing.php
+++ b/web/app/mu-plugins/disallow-indexing.php
@@ -14,11 +14,9 @@ if (!defined('WP_ENV') || WP_ENV === 'production') {
     return;
 }
 
-if (!is_admin()) {
-    add_action('pre_option_blog_public', '__return_zero');
-} else {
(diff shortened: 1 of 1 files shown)


web/app/mu-plugins/disallow-indexing.php
```diff
@@ -3,12 +3,21 @@
 Plugin Name:  Disallow Indexing
 Plugin URI:   https://roots.io/bedrock/
 Description:  Disallow indexing of your site on non-production environments.
-Version:      1.0.0
+Version:      2.0.0
 Author:       Roots
 Author URI:   https://roots.io/
 License:      MIT License
```
JulienMelissas: Do we want to add the text domain for "roots"? 💋

kalenjohnson: this feature is becoming a beast

knowler:
Tested and it works. 👍

Changed notice to a warning instead of an error + added text domain. Ready to be squashed and merged.

web/app/mu-plugins/disallow-indexing.php
```diff
@@ -3,12 +3,22 @@
...
+if (!defined('WP_ENV') || WP_ENV === 'production') {
+    return;
+}
+
+if (!is_admin()) {
     add_action('pre_option_blog_public', '__return_zero');
+} else {
+    add_action('admin_notices', function () {
+        $message = __('Search engine indexing is currently discouraged.', 'roots');
+        echo sprintf('<div class="notice notice-warning"><p>%s</p></div>', $message);
```
aaemnnosttv: Use `printf()` instead of `echo sprintf()`

web/app/mu-plugins/disallow-indexing.php
```diff
@@ -3,12 +3,22 @@
...
 */
 
-if (defined('WP_ENV') && WP_ENV !== 'production' && !is_admin()) {
+if (!defined('WP_ENV') || WP_ENV === 'production') {
+    return;
+}
+
+if (!is_admin()) {
     add_action('pre_option_blog_public', '__return_zero');
+} else {
```
aaemnnosttv: The `else` is not really necessary as there's no harm in hooking onto an admin hook for a non-admin request. I would personally omit it and save the extra braces + indentation; is this a WP/Roots style convention?

austinpray: :+1:

web/app/mu-plugins/disallow-indexing.php
austinpray: why use either? why not just inline the var?

knowler:
Maybe it could be helpful to append "Current environment is [WP_ENV]." to the notice. It could help identify misteaks.

QWp6t:
"Why?" could link to `admin_url('plugins.php?s=disallow-indexing')` or something similar imo.

knowler:
Bedrock: Search engine indexing has been discouraged because the current environment is productoin.

Based on the feedback above I removed the conditional, used inline vars (thanks @austinpray), and made the message read a bit better.

> Search engine indexing is discouraged [for/by/in/with] the current environment

could be read as a suggestion, when what we are actually doing is informing the user of what is happening.

Also, I decided against the "Why?" link and replaced it with **Bedrock** because:

1. mu-plugins don't have pages, so `admin_url('plugins.php?s=disallow-indexing')` doesn't work.
2. It felt redundant linking to a description that is not so much different than the message itself.
3. The plugin is a part of the functionality that _Bedrock_ provides.
4. There is less to add code-wise. 😛

Unless anyone has any issues with this or wants to refine this more, we are ready to go (again).

web/app/mu-plugins/disallow-indexing.php
```diff
@@ -3,12 +3,20 @@
...
 License:      MIT License
 */
 
-if (defined('WP_ENV') && WP_ENV !== 'production' && !is_admin()) {
-    add_action('pre_option_blog_public', '__return_zero');
+if (!defined('WP_ENV') || WP_ENV === 'production') {
+    return;
 }
+
+add_action('pre_option_blog_public', '__return_zero');
```
QWp6t: I'm pretty sure there needs to be `!is_admin()` check around this to prevent WordPress admin panel from always showing the option as enabled.

web/app/mu-plugins/disallow-indexing.php
```diff
@@ -3,12 +3,20 @@
...
+
+add_action('admin_notices', function () {
+    $message = __('Search engine indexing has been discouraged because the current environment is', 'roots');
+    echo "<div class='notice notice-warning'><p><strong>Bedrock:</strong> {$message} <code>".WP_ENV."</code>.</p></div>";
```
QWp6t:
```php
$message = sprintf(__('%1$s Search engine indexing has been discouraged because the current environment is %2$s.', 'roots'), '<strong>Bedrock:</strong>', '<code>'.WP_ENV.'</code>');
echo "<div class='notice notice-warning'><p>{$message}</p></div>";

**Comment by JulienMelissas** [comment]
I'll weigh in @QWp6t... I liked the "Why?" - especially if it just leads to some documentation.

File: web/app/mu-plugins/disallow-indexing.php [hunk 6]
```diff
@@ -3,12 +3,24 @@
...
+Version:      2.0.0
 Author:       Roots
 Author URI:   https://roots.io/
+Text Domain:  roots
 License:      MIT License
 */
 
-if (defined('WP_ENV') && WP_ENV !== 'production' && !is_admin()) {
-    add_action('pre_option_blog_public', '__return_zero');
+if (!defined('WP_ENV') || WP_ENV === 'production') {
```
austinpray: What if we just did `Config::define('DISALLOW_INDEXING', WP_ENV !== 'production');` in config/application.php

and then 

```php
if (!defined('DISALLOW_INDEXING') || DISALLOW_INDEXING !== true) {
}
```

It makes the code read a bunch better:

Currently:
> If we are on prod don't disallow indexing

Proposed:
> If we aren't on prod disallow indexing

Additionally this makes the functionality 
- super easy to trace via config/application.php
- does not rely on magic string 'production'
- ability to be overridden per environment with new config system

austinpray:
Since we are definitely doing extracting out the plugin autoloader (#299) we should probably extract this guy out as well and then implement my suggestions here https://github.com/roots/bedrock/pull/369#discussion_r209290162

retlehs:
picking this up again - i've made https://github.com/roots/bedrock-disallow-indexing and pushed some initial code with @austinpray's suggestion

i also added a filter to prevent the notice from showing

see #521

PR #370, opened 2018-07-31
Adding extra options to .env

vpillinger:
Allow users to set some vars that would otherwise be impossible to set.


QWp6t:
It also discourages environment parity.

Bedrock is delete key friendly and easily configured. Our defaults are what we recommend and are deliberately opinionated.

PR #371, opened 2018-07-31
Making disallow indexing safer

vpillinger:
#368

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -5,6 +5,7 @@ DB_PASSWORD=database_password
 # Optional variables
 # DB_HOST=localhost
 # DB_PREFIX=wp_
+# DISALLOW_INDEXING=true
 
 WP_ENV=development
 WP_HOME=http://example.com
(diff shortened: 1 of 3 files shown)


MWDelaney:
This is removing basic Bedrock functionality and one of the main advantages to haveing different environments in the first place. I don’t think this is the way we should go here.

vpillinger:
How about an alternative here where we move the configuration into config/environments files.

The main issue that I have with the current implementation is that this particular config is hardcoded to a specific environment where it can't be changed. For example, if someone wanted to create "altproduction" environment that doesn't block indexing then they could not with the current implementation. 

This takes advantage of the environments and their associated configuration files in a more robust manner.

austinpray:
@vpillinger this can potentially be fixed by https://github.com/roots/bedrock/pull/380

You could (with the changes in that PR)

1. add a "INDEXING_DISCOURAGED" environment variable
2. have the mu-plugin respect it 
3. Config::define('INDEXING_DISCOURAGED', true) in application.php
4. Config::define('INDEXING_DISCOURAGED', false) in environments/production.php

Now with this model if you feel like it should be "safer" you could just invert it

- Config::define('INDEXING_DISCOURAGED', false) in application.php
- Config::define('INDEXING_DISCOURAGED', true) in environments/*.php except production.php

This is not really possible with the current config scheme

PR #372, opened 2018-08-01
Discourage environment-specific config

austinpray:
Position environment-specific configuration as 'overrides'. Ideally you
want everything to behave as closely as possible to production and only
deviate when it is absolutely necessary such as a QA or development
environment.

- Increases readability
- Increase parity between environments
- Fail safe instead of fail open with regard to defaults in unknown
environments
- DRY


austinpray:
Oh yeah you can't override constants in php

PR #374, opened 2018-08-01
Unnecessary parens

austinpray:


austinpray:
dude DAMMIT I missed this one https://github.com/roots/bedrock/blob/master/web/app/mu-plugins/bedrock-autoloader.php#L138

PR #376, opened 2018-08-02
Adds roave/security-advisories to composer dev deps

austinpray:
https://github.com/Roave/SecurityAdvisories

image

Imo if we include a code sniffer we should include checks for packages that cause security problems. Let's all try this out and see if it works well?

`composer require --dev roave/security-advisories:dev-master`


austinpray:
Worth noting: https://github.com/wp-cli/wp-cli/blob/ecb9ed7e63475aeb00b317013a2b9c48c1a4c8ab/composer.json#L20

PR #377, opened 2018-08-02
Update to WordPress 4.9.8

knowler:
I thought of adding [this](https://gist.github.com/knowler/41d45ed19812a60453a44cbab1be9964) first, but that'd be evil 😄.

Edit: actually, it looks like they've limited the audience for the callout. It doesn't show up in production.


austinpray:
I need to automate this again so I can steal @knowler's commits

PR #380, opened 2018-08-05
New Bedrock Configuration Model

austinpray:
- [x] tag roots/wp-config 1.0.0 before merge

## Problem

The current Bedrock configuration system does not strictly adhere to one of the most important Roots core values: 

> All environments should match production as close as possible.

Stated another way and scoped specifically for Bedrock:

> Environment configuration values should default to production values and only
> deviate if absolutely necessary.

This is not how bedrock currently works. [the current bedrock configuration process][brconfigcode] described by [the docs][brconfigdocs] in general works like so:

oldconfig

Which seems fine at first. However, upon further examination it falls apart.

[The docs][brconfigdocs] describe this intended usage:

> The environment configs are required **before** the main `application` config so anything in an environment config takes precedence over `application`.
> 
> Note: You can't re-define constants in PHP. So if you have a base setting in `application.php` and want to override it in `production.php` for example, you have a few options:
> 
> * Remove the base option and be sure to define it in every environment it's needed
> * Only define the constant in `application.php` if it isn't already defined.

At best the config process:
1. is not DRY 
2. does not fail safe in situations of an unexpected `WP_ENV` value. 
3. will silently reject configuration values if a user is not clear on how `define()`ing constants in php works
4. is not consistent with our environment parity recommendations

### Not DRY

If you want all environments to match production you would have to copy paste the contents of `environments/production.php` to everywhere except `environments/development.php`. Obvious DRY violations aside, this is error prone. `environments/staging.php` can easily get out of sync with `environments/production.php` and congratulations your staging environment no longer reflects production or vice versa.

### Does Not Fail Safe

You could perhaps `require_once __dir__ . '/production.php';` in every file except development.php. However now the second Bedrock encounters an unexpected `WP_ENV` variable (`staging-experimental`, `productoin`, or `undefined` to name a few): Bedrock will fail open and stuff like `WP_DEBUG_DISPLAY` will default to `true` (!!!). No web framework should behave this way when undefined values make their way into the system.

### Silent Configuration Rejections

In ruby

```
irb(main):001:0> WP_DEBUG_DISPLAY = true
=> true
irb(main):002:0> WP_DEBUG_DISPLAY = false
(irb):2: warning: already initialized constant WP_DEBUG_DISPLAY
(irb):1: warning: previous definition of WP_DEBUG_DISPLAY was here
=> false
```

Clearly I am trying to do something bad and the logs will reflect this.

In PHP

```
php > define('WP_DEBUG_DISPLAY', true);
php > define('WP_DEBUG_DISPLAY', false);
php > var_dump(WP_DEBUG_DISPLAY);
bool(true)
```

Yikes.

### Not Consistent With Roots Recommendations

If we recommend:

> All environments should match production as close as possible.

Then why does a user have to take explicit action to make this happen for every environment they define? Why doesn't it just default to this behavior? In the current model production configuration is "opt-in" instead of "opt-out". This is perpendicular to what we recommend.

## Solution

"Production" configuration should be specified in its entirety in config/application.php and files in `config/environments/*.php` should only be used to override production behavior. For instance in development.php you would want to show error messages and define debug flags and stuff.

To do this we should
1. Collect all the configuration values in application.php as key value pairs in a map
2. Collect all the configuration values in the desired `config/environments/WP_ENV.php` file and merge them with the key value pairs collected in application.php
3. After all configuration values are collected we should apply them to the environment via `define()`

At every point in every step above we need to protect the users from redefining constants.

## Implementation

**NOTE:** Before your eyes glaze over looking at crazy UML diagrams: go read the code. What we are doing here is very simple and takes less than 100 lines of code.

```
Lines of Code (LOC)                              127
Comment Lines of Code (CLOC)                      56 (44.09%)
Non-Comment Lines of Code (NCLOC)                 71 (55.91%)
```

The reason I am going so overboard with explaining the implementation is because I need a second pair of eyes to sign off on the UX I am proposing here. I want to validation on if throwing these exceptions in these places makes sense etc. Also in writing this explanation I have already fixed a couple edgecases.

I have defined these classes:

bedrock-classes

Who's usage looks like:

bedrock-classes-usage

To produce:

newconfig

### Exceptions

I throw a `ConstantAlreadyDefinedException` when a user attempts to enter a situation described in the "Problem" section where PHP would silently reject the configuration value.

I throw a `UndefinedConfigKeyException` when a user attempts to fetch a value with `Config::get(k)` that doesn't exist in `Config::$configMap`. I do this because the control flow is nice and `return null` is not working because null is a viable configuration value. We don't use `Config::get(k)` anywhere but it is a part of the API because I can see users needing it.

## Backwards Compatibility

All of this `Config` class madness is opt-in. Nothing stops a user from just going back to using `define()`. `Config::define()` is just an alternative API that behaves in a safe and sane way with regards to overriding environment configurations.

## Conclusion

I am out of time for today so can't elaborate much further. Interested to discuss this with you guys!

[brconfigdocs]: https://github.com/roots/docs/blob/f7ca25c68e5207cee648630df6cdabb3d3da25c5/bedrock/configuration-files.md
[brconfigcode]: https://github.com/roots/bedrock/tree/0668b29c008154b5c47cd236664722f101567aa5/config
[oldenvconfigs]: https://github.com/roots/bedrock/tree/0668b29c008154b5c47cd236664722f101567aa5/config/environments


lib/Bedrock/ConstantAlreadyDefinedException.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Roots\Bedrock;
+
+/**
+ * Class ConstantAlreadyDefinedException
+ * This should be thrown when a user attempts to define() a constant that has already been defined
+ * @package Roots\Bedrock
+ */
+class ConstantAlreadyDefinedException extends \RuntimeException
```
TangRufus: Should we use `LogicException` instead?

Because `LogicException` is
> Exception that represents error in the program logic. This kind of exception should **lead directly to a fix in your code**. 
>
> https://secure.php.net/manual/en/class.logicexception.php

Same goes to `UndefinedConfigKeyException`

austinpray: Hey! Thank you for taking a look at this. 

I understand where you are coming from suggesting when suggesting \LogicException. However, I am still going to advocate for \RuntimeException.

In general when trying to determine what exception class to extend I always look what other exceptions subclass it. Let's look at the exception hierarchy with some cherry-picked examples:

- LogicException (extends Exception): Exception that represents error in the program logic. This kind of exception should lead directly to a fix in your code.
    - BadFunctionCallException: 
    - BadMethodCallException
    - DomainException
    - InvalidArgumentException
    - LengthException
    - OutOfRangeException: Exception thrown when an illegal index was requested. **This represents errors that *should be* detected at compile time**.
- RuntimeException (extends Exception): Exception thrown if **an error which can only be found on runtime occurs**.
    - OutOfBoundsException: Exception thrown if a value is not a valid key. **This represents errors that *cannot be* detected at compile time**.
    - OverflowException
    - RangeException
    - UnderflowException
    - UnexpectedValueException

The fact that the runtime has been polluted with unexpected constants is something that cannot be detected at compile time. Especially since we dynamically build this configuration based on environment variables.

Take a look at the get method of \DS\Map: https://secure.php.net/manual/en/ds-map.get.php. This throws an OutOfBoundsException (extends RuntimeException) if the key could not be found and a default value was not provided.

Lastly consider the following completely valid use case:

JankyHomebrewDBClient.php
```
if (!defined('DB_PORT')) {
  define('DB_PORT', 3306)
}

class JankyHomebrewDBClient {
  // ...
}
```

config/application.php
```
// ...
Config::define('DB_PORT', env('DB_PORT'));
// ...
Config::apply();
```

one-off-db-migration-script.php
```php
// I just want to connect to the DB I don't need WordPress
require_once __DIR__ . '/vendor/autoload.php';
require_once __DIR__ . '/config/application.php';

$client = new JankyHomebrewDBClient(/* ... */);
```

This will throw a `ConstantAlreadyDefinedException` at runtime. I don't necessarily consider this a logical error, since JankyHomebrewDBClient can never tell if it is going to define DB_PORT at compile time.

What do you think?

lib/Bedrock/Config.php
```diff
@@ -0,0 +1,108 @@
...
+    
+    /**
+     * @param string $key
+     * @return bool
+     * @throws ConstantAlreadyDefinedException
+     */
+    private static function defined($key)
+    {
+        if (defined($key)) {
+            $message = "Bedrock aborted trying to redefine constant '$key'";
```
swalkinshaw: Hmm I think we can be a little more clear about what's going on here. Something like:

> Config aborted trying to `define` '$key' . The constant has already been defined.

config/application.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+use Roots\Bedrock\Config;
+
```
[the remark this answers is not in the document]
austinpray: Yep this needs to be in here

lib/Bedrock/Config.php
```diff
@@ -0,0 +1,108 @@
...
+            throw new ConstantAlreadyDefinedException($message);
```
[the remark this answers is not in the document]
austinpray: The user should be able to recover from the exception without incurring a warning though. Lemme think about this

lib/Bedrock/Config.php
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * Class Config
+ * @package Roots\Bedrock
+ */
+class Config
+{
+    /**
+     * @var array
+     */
+    private static $configMap = [];
```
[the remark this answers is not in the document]
austinpray: :+1: changed over at https://github.com/roots/wp-config

austinpray:
Moving this over to https://github.com/roots/wp-config

lib/Bedrock/ConstantAlreadyDefinedException.php
TangRufus: Awesome! This reply should go into textbooks as "best code review reply example".
Thanks!!!

composer.json
```diff
@@ -32,13 +32,19 @@
       "url": "https://wpackagist.org"
     }
   ],
+  "autoload": {
+    "psr-4": {
+      "Roots\\Bedrock\\": "lib/Bedrock"
```
aaemnnosttv: This shouldn't be necessary anymore now that Config has been moved to a package.

composer.json
```diff
@@ -32,13 +32,19 @@
...
+  },
   "require": {
     "php": ">=5.6",
     "composer/installers": "^1.4",
     "vlucas/phpdotenv": "^2.0.1",
     "johnpbloch/wordpress": "4.9.8",
     "oscarotero/env": "^1.1.0",
-    "roots/wp-password-bcrypt": "1.0.0"
+    "roots/wp-password-bcrypt": "1.0.0",
+    "roots/wp-config": "dev-master"
```
aaemnnosttv: +1 for extracting this as a package!

How about `^1.0` instead of `dev-master`?

austinpray: Yep! I am going tag 1.0.0 before this is merged

composer.json
austinpray: whoops! I'm sloppy haha

swalkinshaw:
@austinpray not a bad idea. Let's consider/do that separately.

mike-source:
Has this change been reflected in the docs? 

https://roots.io/bedrock/docs/folder-structure/
https://roots.io/bedrock/docs/configuration-files/

Not sure if this is the best place to mention it, and [probably bad practice on my part but I was copying in an old composer.json (from before August 2018, with my preferred packages) and missed the changes. 

Just posting here in case anyone else is doing the same (I suspect quite a few people do this with bedrock + composer) and got confused. Was pulling my hair out for a couple hours!

austinpray:
Oh not yet 

https://github.com/roots/docs/blob/docs/bedrock/folder-structure.md
https://github.com/roots/docs/blob/docs/bedrock/configuration-files.md

mike-source:
I'd offer to do it myself, but will confess I haven't got my head round the changes yet!

TangRufus:
> probably bad practice on my part but I was copying in an old composer.json (from before August 2018, with my preferred packages) and missed the changes.

A simpler solution: Don't copy and paste the JSON; Use `$ composer require` instead.

mike-source:
@TangRufus hmmm maybe I could write myself a CLI script in future, I needed to add 8+ plugins to my composer.json at once and also add a couple of custom  repositories... ACF pro is incredibly annoying to use via composer.

I'll probably just remember to check Bedrock latest changes!

austinpray:
I can walk y'all boyos through the changes if you wanna contribute docs. Would be more fun that way

MikeiLL:
I'd like to contribute, but am also a little glassy-eyed about where this codebase goes. Is it a plugin that would potentially be installed with composer? Would that be `$ composer require roots/wp-config`?

austinpray:
@MikeiLL (and anyone else reading this) wanna hop on a google hangout later today and I can walk ya through it in like 10 mins? Can record and then share here for posterity

austinpray:
https://doodle.com/poll/mxa2ma2cmdef9aib If y’all want to vote on a time that works

mike-source:
@austinpray I'm UK based so those times are a bit tricky (been a long day!) If something happens definitely share it though I'll take a look. Cheers.

MikeiLL:
@austinpray yes see you then.

austinpray:
Sweet! Just gonna do a hangout on air here https://youtu.be/SNODXZbh8uM so it's recorded for later

austinpray:
https://hangouts.google.com/call/4noypIUSvl1ljNNgs47mAAEE

PR #382, opened 2018-08-23
Updates

k-k-k-kirill:
Updated wordpress and seravo versions to latest


greatislander:
@k-k-k-kirill I think this PR should be against a different repo.

PR #383, opened 2018-09-06
Allow defining WP_CONTENT_URL in .env file

lstellway:
This is helpful when creating a cookie-free domain to serve static content


swalkinshaw:
This seems fine to me... but can you explain the use case about the cookie free domain? Not sure I get it

TangRufus:
Maybe we should add a inline comment to tell users to define `COOKIE_DOMAIN` as well?

See: https://codex.wordpress.org/Editing_wp-config.php#Set_Cookie_Domain

PR #384, opened 2018-09-10
Add a suggestion to the README for Trellis

partounian:
Reading this page along with its docs and Trellis's, I wasn't 100% sure as to whether it made sense to use the `.env` file or not.


swalkinshaw:
Sorry I missed this as the time. But I think this is a better fit for a page in our official docs. If you still feel it's worthwhile, can you propose a change on the docs repo?

PR #405, opened 2018-12-19
Bump requirements to PHP 7.1

retlehs:
PHP 5.6 and 7.0 are no longer supported

http://php.net/supported-versions.php

ref https://twitter.com/rootswp/status/1068635955936907265
ref #267


drzraf:
While I understand the reasoning and votes behind this merge, I think its a bad decision.

**Debian**, for the first time in its history does not provide a version of PHP supported by upstream. But, by definition, provides (security) support for the software bundled (PHP being one of them).

For that reason, even if upstream dropped support `<7.1` I believe distribution must be fairly considered too.
**Debian** does **support** PHP 7.0, but none of the above version are available in Debian stable.

Since almost no one run PHP binary outside a **Linux distribution** I believe the PHP requirement should be set to `>= 7.0` until this version get obsoleted (not as in "upstream speech" but as in "people uses and server boxes")

PR #413, opened 2019-01-30
Detect HTTPS if wordpress is behind a reverse proxy

tristanbes:
Hello,

It'd be great if HTTPS would be working behind a reverse proxy, which is very common when using a PaaS or a WAF, or any decent cloud providers.

https://codex.wordpress.org/Function_Reference/is_ssl#Notes

Since it's a condition, it does not harm existing wordpress installations.

### DIFF
--- a/config/application.php
+++ b/config/application.php
@@ -95,6 +95,10 @@ if (file_exists($env_config)) {
     require_once $env_config;
 }
 
+if (isset($_SERVER['HTTP_X_FORWARDED_PROTO']) && $_SERVER['HTTP_X_FORWARDED_PROTO'] == 'https') {
+    $_SERVER['HTTPS'] = 'on';
+}
+
 Config::apply();
 
 /**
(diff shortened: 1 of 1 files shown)


swalkinshaw:
Seems like a good thing to support by default 👍 

Maybe this should go above loading the env specific config? It would allow them have logic based on this.

config/application.php
```diff
@@ -95,6 +95,10 @@
     require_once $env_config;
 }
 
+if (isset($_SERVER['HTTP_X_FORWARDED_PROTO']) && $_SERVER['HTTP_X_FORWARDED_PROTO'] == 'https') {
```
swalkinshaw: Might be useful to provide a comment referencing https://codex.wordpress.org/Function_Reference/is_ssl#Notes as well

tristanbes:
@swalkinshaw done.

config/application.php
```diff
@@ -89,6 +89,14 @@
 Config::define('SCRIPT_DEBUG', false);
 ini_set('display_errors', 0);
 
+/**
+ * Allow Wordpress to detect HTTPS when used behind a reverse proxy or a load balancer
```
JulienMelissas: Hey there, can we correct to "WordPress" please? https://codex.wordpress.org/Function_Reference/capital_P_dangit 😝

tristanbes: done

swalkinshaw:
Thank you!

PR #412, opened 2019-01-28
Update vlucas/phpdotenv to ^3

knowler:
Resolves #411 

For compatibility with ACF Pro Installer, I made a PR to bump the version requirement over there as well: https://github.com/PhilippBaschke/acf-pro-installer/pull/35

PR #414, opened 2019-01-30
Allow to use a database DSN instead of all the database variables

tristanbes:
Hello,

As mentionned on the .README, 
> Bedrock is inspired by the [Twelve-Factor App](http://12factor.net/) methodology

So I was surprised, during my 120+ Wordpress migration plan to add the same snippet over and over again to allow to use a DSN instead of all the random legacy Wordpress variables (`DB_NAME`, `DB_USER`...).

The 12Factor app is clearly in favor of using DSN: https://12factor.net/backing-services.

Also, when using a recent cloud provider, or PaaS provider, the often expose the database through a DSN (since they follow 12Factor App methodology too).

For those reading the PR and who don't know what a [DSN](https://en.wikipedia.org/wiki/Data_source_name) is, it's a string containing the connection information. Ex: 

```
mysql://database_user:database_password@database_host:database_port/database_name
```

So instead of breaking everything and force people to use DSN instead of legacy variables, this PR adds the support of DSN without breaking existing applications.

Basically, it you set a `DATABASE_URL` variable, then the code takes care of parsing the URL to set the correct variables.

I've been running tests on local (where `DATABASE_URL` was not set) and on production (where `DATABASE_URL` was set) with no issues.

I've updated the example and the documentation too.

Tell me what you think about it.

The algorithm takes care of the following DSN variations:
```
mysql://john_doe:password@127.0.0.1:3402/my_awesome_blog
mysql://john_doe:password@127.0.0.1/my_awesome_blog
mysql://john_doe@127.0.0.1/my_awesome_blog
```

### DIFF
--- a/config/application.php
+++ b/config/application.php
@@ -62,7 +62,7 @@ if (true === $hasDsn) {
 
 Config::define('DB_NAME', $hasDsn ? $dbName : env('DB_NAME'));
 Config::define('DB_USER', $hasDsn ? $databaseDsn['user'] : env('DB_USER'));
-Config::define('DB_PASSWORD', $hasDsn ? $databaseDsn['pass'] : env('DB_PASSWORD'));
+Config::define('DB_PASSWORD', $hasDsn ? ($databaseDsn['pass'] ?? null) : env('DB_PASSWORD'));
 Config::define('DB_HOST', $hasDsn ? $dbHost : env('DB_HOST') ?: 'localhost');
 Config::define('DB_CHARSET', 'utf8mb4');
 Config::define('DB_COLLATE', '');
--- a/config/application.php
+++ b/config/application.php
@@ -57,7 +57,7 @@ $hasDsn = env('DATABASE_URL') !== null ? true : false;
 if (true === $hasDsn) {
     $databaseDsn = parse_url(env('DATABASE_URL'));
     $dbName = substr($databaseDsn['path'], 1);
-    $dbHost = $databaseDsn['host'] . ":" . $databaseDsn['port'];
+    $dbHost = isset($databaseDsn['port']) ? $databaseDsn['host'] . ":" . $databaseDsn['port'] : $databaseDsn['host'];
 }
 
 Config::define('DB_NAME', $hasDsn ? $dbName : env('DB_NAME'));
--- a/config/application.php
+++ b/config/application.php
@@ -64,14 +64,14 @@ Config::define('DB_COLLATE', '');
 $table_prefix = env('DB_PREFIX') ?: 'wp_';
 
 if (env('DATABASE_URL')) {
-    $databaseDsn = parse_url(env('DATABASE_URL'));
-    $dbName = substr($databaseDsn['path'], 1);
-    $dbHost = isset($databaseDsn['port']) ? $databaseDsn['host'] . ":" . $databaseDsn['port'] : $databaseDsn['host'];
+    $database_dsn = parse_url(env('DATABASE_URL'));
+    $db_name = substr($database_dsn['path'], 1);
+    $db_host = isset($database_dsn['port']) ? $database_dsn['host'] . ":" . $database_dsn['port'] : $database_dsn['host'];
 
-    Config::define('DB_NAME', $dbName);
-    Config::define('DB_USER', $databaseDsn['user']);
-    Config::define('DB_PASSWORD', $databaseDsn['pass'] ?? null);
-    Config::define('DB_HOST', $dbHost);
+    Config::define('DB_NAME', $db_name);
+    Config::define('DB_USER', $database_dsn['user']);
+    Config::define('DB_PASSWORD', $database_dsn['pass'] ?? null);
+    Config::define('DB_HOST', $db_host);
 }
 
 /**
--- a/config/application.php
+++ b/config/application.php
@@ -29,7 +29,7 @@ if (file_exists($root_dir . '/.env')) {
     $dotenv->load();
     $dotenv->required(['WP_HOME', 'WP_SITEURL']);
     if (!env('DATABASE_URL')) {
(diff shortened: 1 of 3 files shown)


swalkinshaw:
I'm not necessarily against this, but wanted to point out two things:

1. that 12 factor page doesn't strictly recommend DSNs. Their image happens to use them but that doesn't mean much.
> To the app, both are attached resources, accessed via a URL or other locator/credentials stored in the config. A deploy of the twelve-factor app should be able to swap out a local MySQL database with one managed by a third party (such as Amazon RDS) without any changes to the app’s code.

"other locator/credentials stored in the config" is the key part which is exactly what Bedrock is doing.

2. the reason why we haven't supported the DSN option is because WordPress itself doesn't. This means we need to do URL parsing like this PR does.

I'm on the fence about this but open to opinions.

Is this something you'd consider trying to contribute to WP core? It should definitely be supported by default but contributing to WP isn't something I'd wish on anyone 😂

tristanbes:
`the reason why we haven't supported the DSN option is because WordPress itself doesn't.`

They also doesn't support your structure, doesn't support composer, and dotenv. Your edition is "the right way" of doing Wordpress (yes Wordpress, we're not in 2004 anymore). So adding support is one step closer your goal. Also, it doesn't remove the actual way of doing :-), it just remove a little pain for cloud/paas users of bedrock

Also: Nope, I don't want to contribute to Wordpress, they are decades away of good practices when it comes to development. I imagine PR like this would be left over like the `composer` support... 😄

tristanbes:
Also, I fixed a case where the code couldn't support a passwordless database connexion. Good catch @Kocal.

tristanbes:
The algorithm takes care of the following DSN variations:
```
mysql://john_doe:password@127.0.0.1:3402/my_awesome_blog
mysql://john_doe:password@127.0.0.1/my_awesome_blog
mysql://john_doe@127.0.0.1/my_awesome_blog
```

I think this PR is now ready if you don't have more comments to add.

config/application.php
```diff
@@ -52,10 +52,18 @@
 /**
  * DB settings
  */
-Config::define('DB_NAME', env('DB_NAME'));
-Config::define('DB_USER', env('DB_USER'));
-Config::define('DB_PASSWORD', env('DB_PASSWORD'));
-Config::define('DB_HOST', env('DB_HOST') ?: 'localhost');
+$hasDsn = env('DATABASE_URL') !== null ? true : false;
```
QWp6t: This could be reduced to:
```
$hasDsn = env('DATABASE_URL') !== null;
```
But the line isn't necessary, so I'd suggest removing it. There's no need to use the `$hasDsn` variable.

config/application.php
```diff
@@ -52,10 +52,18 @@
...
+
+if (true === $hasDsn) {
```
QWp6t: You can use `if (env('DATABASE_URL')) {` instead.

config/application.php
```diff
@@ -52,10 +52,18 @@
...
+    $databaseDsn = parse_url(env('DATABASE_URL'));
+    $dbName = substr($databaseDsn['path'], 1);
+    $dbHost = isset($databaseDsn['port']) ? $databaseDsn['host'] . ":" . $databaseDsn['port'] : $databaseDsn['host'];
+}
+
+Config::define('DB_NAME', $hasDsn ? $dbName : env('DB_NAME'));
```
QWp6t: These `$hasDsn` ternary operations aren't needed. `Config::define()` can be declared more than once or you can set the env var.

In other words, you can use
```php
if (env('DB_URL')) {
    //...
    putenv('DB_...');// or $_ENV['DB_...'] = ...
    //...
}

Config::define('DB_...');
```

Or you could do the following...

```php
Config::define('DB_...');
// ...

if (env('DB_URL')) {
  Config::define('DB_...');
  //...
}
```

I personally prefer the latter.

tristanbes:
@QWp6t do you find this version better ?

config/application.php
```diff
@@ -60,6 +60,17 @@
 Config::define('DB_COLLATE', '');
 $table_prefix = env('DB_PREFIX') ?: 'wp_';
 
+if (env('DATABASE_URL')) {
```
swalkinshaw: 🤔 we require `DB_NAME', 'DB_USER', 'DB_PASSWORD'` when .env is loaded. We'd need to make that optional somehow wouldn't we?

tristanbes:
What is missing in order to merge and tag a new release of bedrock ? :D

swalkinshaw:
My question about the required env vars is still valid I think. If you just use `DATABASE_URL` right now, wouldn't this fail validation?

tristanbes:
@swalkinshaw I've updated the code to allow only `DATABASE_URL` to be defined

austinpray:
I'll give this a go tomorrow and merge.

tristanbes:
Great news thanks :-), if possible tag a new release right away so we can all benefit from my 2 last PR 🤓

config/application.php
```diff
@@ -60,6 +63,17 @@
...
+    $databaseDsn = parse_url(env('DATABASE_URL'));
```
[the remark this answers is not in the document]
tristanbes: And done. Sorry, I follow PSR standard so I naturally used camelCase variables

tristanbes:
Is there something missing I can add that prevents this PR to get merged ? :)

austinpray:
Sweet, works here!

austinpray:
> Great news thanks :-), if possible tag a new release right away so we can all benefit from my 2 last P

@swalkinshaw @retlehs Let's get #412 in and then tag a new release? 1.12.0?

tristanbes:
Thank you, @austinpray thanks for providing reactive feedback on my work

PR #415, opened 2019-02-08
Make DSN implementation more uniform

Log1x:
Figured I'd propose this change as the implementation is still early so despite being a "breaking change" – very few would be affected.

- Change `DATABASE_URL` to `DB_URL` to remain uniform with the other database variables
- Improve phrasing on the README and `.env.example` DocBlock
- Minor golfing / cleanup on how the DSN is handled in `application.php`
- Apparently editorconfig found some trailing whitespace in the README 🙀

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -3,8 +3,8 @@ DB_USER=database_user
 DB_PASSWORD=database_password
 
 # Optionally, you can use a data source name (DSN)
-# When using a DSN, you can leave DB_NAME, DB_USER, DB_PASSWORD, and DB_HOST empty
-# DB_URL=mysql://database_user:database_password@database_host:database_port/database_name
+# When using a DSN, you can remove the DB_NAME, DB_USER, DB_PASSWORD, and DB_HOST variables
+# DATABASE_URL=mysql://database_user:database_password@database_host:database_port/database_name
 
 # Optional variables
 # DB_HOST=localhost
--- a/.env.example
+++ b/.env.example
@@ -2,9 +2,9 @@ DB_NAME=database_name
 DB_USER=database_user
 DB_PASSWORD=database_password
 
-# Or you can use a DSN (database source name) 
-# In this case you can leave empty variables DB_NAME, DB_USER, DB_PASSWORD, DB_HOST 
-# DATABASE_URL=mysql://database_user:database_password@database_host:database_port/database_name
+# Optionally, you can use a data source name (DSN)
+# When using a DSN, you can leave DB_NAME, DB_USER, DB_PASSWORD, and DB_HOST empty
+# DB_URL=mysql://database_user:database_password@database_host:database_port/database_name
 
 # Optional variables
 # DB_HOST=localhost
--- a/config/application.php
+++ b/config/application.php
@@ -69,7 +69,7 @@ if (env('DATABASE_URL')) {
     Config::define('DB_NAME', substr($dsn->path, 1));
     Config::define('DB_USER', $dsn->user);
     Config::define('DB_PASSWORD', $dsn->pass ?? null);
-    Config::define('DB_HOST', isset($dsn->port) ? $dsn->host . ':' . $dsn->port : $dsn->host);
+    Config::define('DB_HOST', isset($dsn->port) ? "$dsn->host:$dsn->port" : $dsn->host);
 }
 
 /**
--- a/config/application.php
+++ b/config/application.php
@@ -69,7 +69,7 @@ if (env('DATABASE_URL')) {
     Config::define('DB_NAME', substr($dsn->path, 1));
     Config::define('DB_USER', $dsn->user);
     Config::define('DB_PASSWORD', $dsn->pass ?? null);
-    Config::define('DB_HOST', isset($dsn->port) ? "$dsn->host:$dsn->port" : $dsn->host);
+    Config::define('DB_HOST', isset($dsn->port) ? "{$dsn->host}:{$dsn->port}" : $dsn->host);
 }
 
 /**
--- a/config/application.php
+++ b/config/application.php
@@ -28,7 +28,7 @@ $dotenv = Dotenv\Dotenv::create($root_dir);
 if (file_exists($root_dir . '/.env')) {
     $dotenv->load();
     $dotenv->required(['WP_HOME', 'WP_SITEURL']);
-    if (!env('DB_URL')) {
+    if (!env('DATABASE_URL')) {
         $dotenv->required(['DB_NAME', 'DB_USER', 'DB_PASSWORD']);
     }
 }
@@ -55,16 +55,16 @@ Config::define('WP_CONTENT_URL', Config::get('WP_HOME') . Config::get('CONTENT_D
 /**
  * DB settings
  */
-Config::define('DB_NAME', env('DB_NAME'));
-Config::define('DB_USER', env('DB_USER'));
-Config::define('DB_PASSWORD', env('DB_PASSWORD'));
+Config::define('DB_NAME', env('DB_NAME') ?? null);
+Config::define('DB_USER', env('DB_USER') ?? null);
+Config::define('DB_PASSWORD', env('DB_PASSWORD') ?? null);
 Config::define('DB_HOST', env('DB_HOST') ?: 'localhost');
 Config::define('DB_CHARSET', 'utf8mb4');
 Config::define('DB_COLLATE', '');
 $table_prefix = env('DB_PREFIX') ?: 'wp_';
 
-if (env('DB_URL')) {
-    $dsn = (object) parse_url(env('DB_URL'));
+if (env('DATABASE_URL')) {
+    $dsn = (object) parse_url(env('DATABASE_URL'));
 
     Config::define('DB_NAME', substr($dsn->path, 1));
     Config::define('DB_USER', $dsn->user);
--- a/config/application.php
+++ b/config/application.php
@@ -28,7 +28,7 @@ $dotenv = Dotenv\Dotenv::create($root_dir);
 if (file_exists($root_dir . '/.env')) {
     $dotenv->load();
     $dotenv->required(['WP_HOME', 'WP_SITEURL']);
-    if (!env('DATABASE_URL')) {
+    if (!env('DB_URL')) {
         $dotenv->required(['DB_NAME', 'DB_USER', 'DB_PASSWORD']);
     }
 }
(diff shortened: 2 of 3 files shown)


swalkinshaw:
While `DB_URL` is more consistent with the other constants, `DATABASE_URL` is actually kind of standard. Lots of hosting platforms use that name specifically and some other web frameworks have it integrated as well.

All the other changes look good though 👍

Log1x:
Want to wait for more feedback or should I change it?

Drupal (lol) uses `DB_URL` – Laravel, etc. doesn't use DSN which isn't a surprise, it isn't user-friendly and there is no definitive standard (e.g. PDO: `mysql:host=localhost;dbname=example`, Perl: `DBI:mysql:database=$database;host=$hostname;port=$port`).

Otherwise, I mainly just see Rails/Django defining `DATABASE_URL`.

swalkinshaw:
Heroku and Doccu use `DATABASE_URL`. There's probably other hosts/infra related things which do as well.

tristanbes:
All the suggested changes looks good except changing `DATABASE_URL` to `DB_URL`

I would stick to `DATABASE_URL` since a lot of cloud providers are using this to expose their DSN. (Scalingo uses it, clever-cloud uses it and other international ones like @swalkinshaw reported).

.env.example
```diff
@@ -2,9 +2,9 @@ DB_NAME=database_name
 DB_USER=database_user
 DB_PASSWORD=database_password
 
-# Or you can use a DSN (database source name) 
-# In this case you can leave empty variables DB_NAME, DB_USER, DB_PASSWORD, DB_HOST 
-# DATABASE_URL=mysql://database_user:database_password@database_host:database_port/database_name
+# Optionally, you can use a data source name (DSN)
+# When using a DSN, you can leave DB_NAME, DB_USER, DB_PASSWORD, and DB_HOST empty
```
tristanbes: Actually, with my latest changes, you can change to: 
```suggestion
# When using a DSN, you can remove DB_NAME, DB_USER, DB_PASSWORD, and DB_HOST variables
```

.env.example
```diff
@@ -2,9 +2,9 @@ DB_NAME=database_name
...
+# DB_URL=mysql://database_user:database_password@database_host:database_port/database_name
```
tristanbes: 👎 on this, see my comment on the PR

config/application.php
```diff
@@ -63,15 +63,13 @@
 Config::define('DB_COLLATE', '');
 $table_prefix = env('DB_PREFIX') ?: 'wp_';
 
-if (env('DATABASE_URL')) {
-    $database_dsn = parse_url(env('DATABASE_URL'));
-    $db_name = substr($database_dsn['path'], 1);
-    $db_host = isset($database_dsn['port']) ? $database_dsn['host'] . ":" . $database_dsn['port'] : $database_dsn['host'];
+if (env('DB_URL')) {
+    $dsn = (object) parse_url(env('DB_URL'));
```
tristanbes: I don't see the point of using implicit cast from an array to an object. Plus, I don't know if it's a good practice to do that 🤔

.env.example
Log1x: With the current implementation, the existence of the DB_ variables isn't checked so removing them would throw an error. We would need to add a null coalesce operator.

config/application.php
tristanbes: Also, it add negative performance to do such casting.
We talking about microseconds, but, keep in mind that application.php is called upon *every* request.
So that microseconds penalty will apply to all requests. And I personally want to avoid that when/if possible.

austinpray:
+1 for DATABASE_URL

config/application.php
```diff
@@ -55,23 +55,21 @@
...
-    Config::define('DB_NAME', $db_name);
-    Config::define('DB_USER', $database_dsn['user']);
-    Config::define('DB_PASSWORD', $database_dsn['pass'] ?? null);
-    Config::define('DB_HOST', $db_host);
+    $dsn = (object) parse_url(env('DATABASE_URL'));
+
+    Config::define('DB_NAME', substr($dsn->path, 1));
+    Config::define('DB_USER', $dsn->user);
+    Config::define('DB_PASSWORD', $dsn->pass ?? null);
+    Config::define('DB_HOST', isset($dsn->port) ? $dsn->host . ':' . $dsn->port : $dsn->host);
```
knowler: I find this cleaner:

```suggestion
    Config::define('DB_HOST', isset($dsn->port) ? "$dsn->host:$dsn->port" : $dsn->host);
```

config/application.php
tristanbes: May I have your feedback on the performance problematic here @austinpray && @knowler ?

config/application.php
Log1x: I considered that– it just isn't being used elsewhere.

config/application.php
Log1x: I think looking at it from a performance prospective is incredibly silly. Every database query processor in Laravel casts to objects, not to mention eloquent. In this case, it is a single array. If we were pinching on performance to this extent, dotenv as a whole would be catastrophic, @austinpray's [wp-config](https://github.com/roots/wp-config) would be excessive overhead, heck, even using `parse_url()` is a huge performance impact. According to [this publication](https://books.google.ca/books?id=mCpnlNYzqOQC&pg=PT732&lpg=PT732&dq=is+parse_url+faster+than+preg_match%3F&source=bl&ots=qzZXezlo-6&sig=KX8areZ0ht01SrN3g_PmDi_IZEQ&hl=en&sa=X&ei=syeWUvDgE4zWoASw04KoAQ#v=onepage&q=parse_url&f=false) – doing it with `preg_match()` is 7.5x faster than `parse_url()`.

I think it comes down to readability and what feels more appropriate. I felt like this was a better fit in both cases.

knowler: While, I do like working with objects since they are syntactically nicer to write — especially for string concatenations — I don’t think the pay off is much since we’re only using it for a few lines within an if clause that nobody will ever touch again. In this case, I’d vote performance over syntactical sugar since, like you said, this is called on every request. We should strive to keep the core as fast as possible, even if it’s a matter of microseconds.

knowler: Really good points @Log1x — your comment didn't show up until after I posted mine. Principally, I still hold to what I said.

austinpray: Rofl can't believe I am even doing this

https://3v4l.org/Kkrc9/perf#output vs https://3v4l.org/FlTqR

`$dsn = (object) parse_url('postgres://AAAAAA:BBBBBBBBBBBBBBBBBBBB@CCCCCCCCCCCCCCCCCCCCCCCCCC.compute-1.amazonaws.com:5432/ddddddddd');` vs `$dsn = parse_url('postgres://AAAAAA:BBBBBBBBBBBBBBBBBBBB@CCCCCCCCCCCCCCCCCCCCCCCCCC.compute-1.amazonaws.com:5432/ddddddddd');`

| Trial | PHP Version | User Time (s) | 
| --- | --- | --- |
| (object) | 7.3 | 0.009 |
| (object) | 7.3 | 0.009 |
| (object) | 7.3 | 0.009 |
| no cast | 7.3 | 0.007 |
| no cast | 7.3 | 0.007 |
| no cast | 7.3 | 0.007 |
| (object) | 7.2 | 0.008 |
| (object) | 7.2 | 0.008 |
| (object) | 7.2 | 0.008 |
| no cast | 7.2 | 0.006 |
| no cast | 7.2 | 0.006 |
| no cast | 7.2 | 0.006 |
| (object) | 7.1 | 0.017 |
| (object) | 7.1 | 0.017 
| (object) | 7.1 | 0.017 |
| no cast | 7.1 | 0.003 |
| no cast | 7.1 | 0.003 |
| no cast | 7.1 | 0.003 |

7.1 takes a hit, but that is about to be EOLed this year. Nobody should be starting a new bedrock project on 7.1, why would they?

I say just roll with (object), it looks nicer.

austinpray: @swalkinshaw tie breaker please haha

tristanbes: I do agree that performance impact are very very very tiny, but only for >PHP 7.1 setup. The question was revelant as the code manipulation is only needed there. 

Thanks for your feedback

config/application.php
QWp6t: I vote for `"{$dsn->host}:{$dsn->port}"` 🤷‍♂️

config/application.php
```diff
@@ -55,23 +55,21 @@
 /**
  * DB settings
  */
-Config::define('DB_NAME', env('DB_NAME'));
-Config::define('DB_USER', env('DB_USER'));
-Config::define('DB_PASSWORD', env('DB_PASSWORD'));
+Config::define('DB_NAME', env('DB_NAME') ?? null);
```
QWp6t: `?? null` seems unnecessary.

Log1x: It was done out of @tristanbes's suggestion above to change the Docblock to include `# When using a DSN, you can **remove** DB_NAME, DB_USER, and DB_PASSWORD` instead of leaving it blank. If there isn't an `isset()` on the env variables and you remove them, dotenv will throw an error.

Should I revert it and just assume people won't comment out or remove the DB_ vars?

QWp6t: `env()` isn't part of `DotEnv`; it won't throw an error. I'm pretty sure you should be able to remove `?? null` here and it will work just fine. I believe `env()` will return an empty string if the values are in the `.env` file, otherwise it will return null. As for `DotEnv`, it won't throw an error because it only requires them if `DATABASE_URL` isn't in your `.env` (application.php#L31-33).

Feel free to double-check if you want. Or I can check tomorrow.

Log1x: Yeah you're right. I had tested it without uncommenting `DATABASE_URL` and mistakened the ValidationException. It works fine without `?? null` as long as `DATABASE_URL` is set.
```
Fatal error: Uncaught Dotenv\Exception\ValidationException: One or more environment variables failed assertions: DB_NAME is missing, DB_USER is missing, DB_PASSWORD is missing.
```

QWp6t:
lgtm @swalkinshaw

config/application.php
```diff
@@ -64,14 +64,12 @@
...
-
-    Config::define('DB_NAME', $db_name);
-    Config::define('DB_USER', $database_dsn['user']);
-    Config::define('DB_PASSWORD', $database_dsn['pass'] ?? null);
-    Config::define('DB_HOST', $db_host);
+    $dsn = (object) parse_url(env('DATABASE_URL'));
+
+    Config::define('DB_NAME', substr($dsn->path, 1));
+    Config::define('DB_USER', $dsn->user);
+    Config::define('DB_PASSWORD', $dsn->pass ?? null);
```
swalkinshaw: @QWp6t didn't you say `?? null` wasn't needed? Or would this result in `""` which would actually be used as the password instead of "skipped"?

Log1x: I had kept it from the original implementation but it does look like it would be fine to remove it as `$dsn->pass` is returning `NULL` and doesn't appear to be throwing any errors when testing with:

```
DATABASE_URL=mysql://root@localhost/wp_sage
```

while having the DB_ vars commented out.

tristanbes: and that's why i'm not a fan of forcing an array into an object via casting. Now you're accessing properties that may not exist in the case you posted in your comment and it will trigger warnings.

```
DATABASE_URL=mysql://root@localhost/wp_sage
```

https://3v4l.org/Yf3ck

Working with objects is great but only if the source was crafted/thought as an object ;)

swalkinshaw: Good catch, thank you! So `?? null` does this fix. Should be noted that the array would behave the same anyway? Either way we need `??`

tristanbes: yeah in both cases we need it to handle passwordless db

Log1x: Yeah, I was going to post an update to my above– I didn't push any changes as I assumed there was a "gotcha" somewhere. My PHP wasn't throwing errors locally but I assumed it *should*.

And yes, for the record, an array would behave the same way. https://3v4l.org/LTTEX – it has nothing to do with casting to an object. :)

swalkinshaw:
Thank you everyone

We also need the docs updated at https://github.com/roots/docs/blob/docs/bedrock/installing-bedrock.md if someone wants to help

config/application.php
austinpray: For posterity: 

If you were skeptical about my above table's results: you were correct to think so. The "User Time (s)" is off by a couple orders of magnitude.

Here's a better benchmark: https://gist.github.com/austinpray/ed8e2aab4b757f1e41d5298584eb1590 / https://docs.google.com/spreadsheets/d/1_MAFd_pxs333myM8WPsOnzBb0hh4ZtvJ5r2zj2HAfJw/edit?usp=sharing

image

So yeah there's a difference in php 7.1 but 7.2+ there is literally no difference whatsoever.

PR #421, opened 2019-02-25
Swap Usage of PHP 7 Null Coalesce Operator 

devkinetic:
The DSN support added last week broke all of my sites running Bedrock in PHP5. I see that PHP5 support has been dropped from the project a few weeks ago, but for the sake of argument, the next line below is using syntax that is compatible.


devkinetic:
Thanks guys! I appreciate it. It would be nice if the readme explicitly stated the last officially supported 5.6 release.

PR #423, opened 2019-02-26
Adjust logic to handle empty DB_PREFIX

devkinetic:
I had a site that was not using prefixes on the database. Bedrock does not support this and evaluates `DB_PREFIX=''` as `false`.

### DIFF
--- a/config/application.php
+++ b/config/application.php
@@ -61,7 +61,7 @@ Config::define('DB_PASSWORD', env('DB_PASSWORD'));
 Config::define('DB_HOST', env('DB_HOST') ?: 'localhost');
 Config::define('DB_CHARSET', 'utf8mb4');
 Config::define('DB_COLLATE', '');
-$table_prefix = (env('DB_PREFIX') || env('DB_PREFIX') === '') ? env('DB_PREFIX') : 'wp_';
+$table_prefix = empty(env('DB_PREFIX')) ? env('DB_PREFIX') : 'wp_'
 
 if (env('DATABASE_URL')) {
     $dsn = (object) parse_url(env('DATABASE_URL'));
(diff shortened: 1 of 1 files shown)


config/application.php
```diff
@@ -61,7 +61,7 @@
 Config::define('DB_HOST', env('DB_HOST') ?: 'localhost');
 Config::define('DB_CHARSET', 'utf8mb4');
 Config::define('DB_COLLATE', '');
-$table_prefix = env('DB_PREFIX') ?: 'wp_';
+$table_prefix = (env('DB_PREFIX') || env('DB_PREFIX') === '') ? env('DB_PREFIX') : 'wp_';
```
swalkinshaw: What about just `empty(env('DB_PREFIX')) ? env('DB_PREFIX') : 'wp_';`

devkinetic: Looks good to me, I'll get this into a patch tomorrow morning so this is wrapped up.

swalkinshaw:
Thank you!

sandrowuermli:
`$table_prefix = ! empty(env('DB_PREFIX')) ? env('DB_PREFIX') : 'wp_'`
or
`$table_prefix = empty(env('DB_PREFIX')) ? 'wp_' : env('DB_PREFIX')`

But please, don't break everyone's WordPress who doesn't have the default prefix `wp_`.

sandrowuermli:
WordPress Setup does not allow empty prefixes:

```
if ( empty( $prefix ) ) {
    wp_die( __( '<strong>ERROR</strong>: "Table Prefix" must not be empty.' ) . $tryagain_link );
}
```

found in:
`wp-admin/setup-config.php`

So, there's nothing to fix here.

devkinetic:
Setup might not. This was from an import of an existing site, gotta wonder how it got like that. What seems better would be to rename the tables to have a prefix.

PR #427, opened 2019-04-02
Configure wp-cli so that `wp server` uses correct webroot

calind:
This configures `wp server` to use the proper webroot by default, so that you don't need to run `wp server --docroot=web`.


swalkinshaw:
Good idea 👍

PR #428, opened 2019-04-13
If WP_HOME it's defined in config file, domain is not set in WP_CONTENT_URL

nopticon:
When you use config files instead of .env files, when defining WP_CONTENT_URL, Config::get('WP_HOME') is empty, so WP_CONTENT_URL value is set as:
`/app` instead of `<WP_HOME value>/app`

This happens when you create `config/environments/<file>.php` and has defined `Config::define('WP_HOME', 'http://example.com');`.

### DIFF
--- a/config/application.php
+++ b/config/application.php
@@ -50,7 +50,6 @@ Config::define('WP_SITEURL', env('WP_SITEURL'));
  */
 Config::define('CONTENT_DIR', '/app');
 Config::define('WP_CONTENT_DIR', $webroot_dir . Config::get('CONTENT_DIR'));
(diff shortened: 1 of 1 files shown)


swalkinshaw:
🤔 while this makes some sense for `WP_CONTENT_URL` specifically, it breaks our principles of environment config files taking precedence over the defaults in `application`.

This means if someone does `Config::define('WP_CONTENT_URL', 'whatever')` in an environment file, it will be overwritten later.

Can't you just define `WP_CONTENT_URL` explicitly instead of only `WP_HOME`?

austinpray:
Ditto what @swalkinshaw says, this makes it impossible to override WP_CONTENT_URL

nopticon:
Yes, you are right. It's better to define WP_CONTENT_URL in config file. Thank you!

PR #437, opened 2019-06-08
Remove unnecessary static variables

nasyrov:
Since the autoloader is singeleton there is no reason to keep the static variables

### DIFF
--- a/web/app/mu-plugins/bedrock-autoloader.php
+++ b/web/app/mu-plugins/bedrock-autoloader.php
@@ -30,10 +30,10 @@ class Autoloader
     private $cache;
 
     /** @var array Autoloaded plugins */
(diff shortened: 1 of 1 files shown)


austinpray:
Haven’t tested this yet but this is a good update 👍

swalkinshaw:
Ditto 👍

nasyrov:
@austinpray, @swalkinshaw Cheers guys!

I was also thinking, shall I update all the variable names to follow PSR2? Just to be consistent with the code.

swalkinshaw:
Yeah good catch. Might as well 😄

nasyrov:
@swalkinshaw this is done, could you please have a look?

swalkinshaw:
Thank you 👏

PR #440, opened 2019-06-19
Ignore debug.log

peterpalau:
This add `debug.log` file to `.gitignore`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,7 @@ web/wp
 web/.htaccess
 
 # Logs
-web/app/debug.log
+*.log
 
 # Dotenv
 .env
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,9 @@ web/app/uploads/*
 web/wp
 web/.htaccess
 
+# Logs
+web/app/debug.log
+
 # Dotenv
 .env
 .env.*


retlehs:
thanks for the PR! we don't define `WP_DEBUG_LOG` in bedrock, and when you define it you're allowed to define a specific location for the logfile..

not sure if this PR should be merged or not. seems like it would make sense if we defined `WP_DEBUG_LOG` w/o a path

swalkinshaw:
But the default is just `define( 'WP_DEBUG_LOG', true );` which would result in this path. So it seems reasonable for Bedrock to ignore it?

peterpalau:
Yes, good point, in that case we don't need right now this modification until you defined `WP_DEBUG_LOG` w/o a path.

peterpalau:
> But the default is just `define( 'WP_DEBUG_LOG', true );` which would result in this path. So it seems reasonable for Bedrock to ignore it?

This is the case I have in the `development` stage.

aaemnnosttv:
Why not ignore `*.log` ? I can't imagine someone would want to include a `.log` file in their repo but if they did, they could always whitelist it.

austinpray:
+1 to just ignoring *.log

I don’t see anything wrong with it and [the community at large seems to be ok with it](https://github.com/github/gitignore/search?utf8=%E2%9C%93&q=%27*.log%27&type=). Popular node gitignore has the same line: https://github.com/github/gitignore/blob/e58b3e0a015871e4caafd2eed04dc243ef5f30b4/Node.gitignore

aaemnnosttv:
WordPress ignore has it as well 😄 

https://github.com/github/gitignore/blob/f9291de89f5f7dc0d3d87f9eb111b839f81d5dbc/WordPress.gitignore#L20-L23

peterpalau:
Yes, they ignore the `debug.log` file by default:

https://core.trac.wordpress.org/browser/branches/5.2/.gitignore#L54

Issue #451, opened 2019-07-26
Modify application.php to support hosts that uses non-standard mysql ports


carl-alberto:
* [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
* [x] This request isn't a duplicate of an existing request
* [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

## Summary
Some hosts uses non-standard mysql ports and there are no way to use it out of the box without modifying the application.php file

## Motivation

This allows users to append a port in the mysql define whenever the port value is present in the .env file

aaemnnosttv:
This seems unnecessary. Can't the port be set as part of `DB_HOST` ?

It can definitely be set using the new `DATABASE_URL` connection string alternative added in #414

PR #452, opened 2019-07-26
Allow custom mysql port to be used whenever that value is present

carl-alberto:
This closes issue #451 

props to @carlalexander based from his Bedrock fork https://github.com/carlalexander/bedrock-pantheon-circleci


swalkinshaw:
Thanks for the contribution but I'm closing this due to https://github.com/roots/bedrock/issues/451#issuecomment-515405133. Feel free to re-open if you feel strongly that this is needed and we can discuss it more.

PR #454, opened 2019-08-03
Various clean up

Log1x:
# Change log

- Clean up linting rules
- Handle rules in `phpcs` CLI instead of a separate `phpcs.xml`
- Fix & clean up various syntax/docblocks to be more compliant with PSR-12
- Add config optimizations to `composer.json`
- Bump dependencies
- Change build badge to CircleCI


composer.json
```diff
@@ -45,6 +42,13 @@
     "squizlabs/php_codesniffer": "^3.4.2",
     "roave/security-advisories": "dev-master"
   },
+  "config": {
+    "optimize-autoloader": true,
+    "preferred-install": "dist",
+    "sort-packages": true
+  },
+  "minimum-stability": "dev",
```
austinpray: What’s up with this?

composer.json
```diff
@@ -58,7 +62,7 @@
       "php -r \"copy('.env.example', '.env');\""
     ],
     "test": [
-      "phpcs"
+      "phpcs --ignore=vendor,web --extensions=php --standard=PSR12 --exclude=PSR1.Files.SideEffects ."
```
austinpray: Revert dis

config/application.php
```diff
@@ -120,6 +130,6 @@
 /**
  * Bootstrap WordPress
  */
-if (!defined('ABSPATH')) {
-    define('ABSPATH', $webroot_dir . '/wp/');
+if (! defined('ABSPATH')) {
```
austinpray: Lol I hate this space after prefixed unary operators. Is this our coding style now?

phpcs.xml
```diff
@@ -1,23 +0,0 @@
-<?xml version="1.0"?>
```
austinpray: Are we ditching this?

Log1x: only if I kept `phpcs --ignore=vendor,web --extensions=php --standard=PSR12 --exclude=PSR1.Files.SideEffects .` which p much does the same thing

config/application.php
Log1x: https://www.php-fig.org/psr/psr-1/#23-side-effects scroll down little bit to conditional example. /shrug – seems to be the norm now

composer.json
Log1x: https://getcomposer.org/doc/articles/autoloader-optimization.md#optimization-level-1-class-map-generation + package sorting when doing `composer require ...` or even `composer update` after adding a package by hand (a-z)

austinpray:
breaking this into multiple PRs

PR #455, opened 2019-08-07
Use inline phpcs rules instead of phpcs.xml

Log1x:
- **Opinionated change**: Lets us take a file out of version control while still maintaining near-identical linting rules.


swalkinshaw:
Agreed this is better 👍

austinpray:
I actually don’t like this change at all. I end up adding a ton of exclusions so third party plugins aren’t scanned. The rules as-is will scan the plugins dir. it doesn’t make sense to add a bunch of those exclusions as cli args in a json file.

TangRufus:
> I actually don’t like this change at all.
Same here.

Having `phpcs.xml` is easier to use custom standards.

austinpray:
Maybe we take this to revert town https://github.com/roots/bedrock/pull/461

PR #459, opened 2019-08-07
Clean up docblocks & commenting

Log1x:
- **Opinionated change**: Clean up existing docblocks as well as make them more uniform across the entire package.

### DIFF
--- a/config/application.php
+++ b/config/application.php
@@ -10,10 +10,18 @@
 
 use Roots\WPConfig\Config;
 
-/** @var string Directory containing all of the site's files */
+/**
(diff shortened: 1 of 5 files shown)


swalkinshaw:
I'm iffy on the vertical aligning as I generally don't like that since they're annoying to modify. However we rarely change this stuff

Log1x:
I feel like Bedrock being the foundation and staple that it is, the code base should be "beautified" and as readable as possible from the end-user– especially as most of it is "dev-facing" (e.g. `application.php`).

PR #461, opened 2019-08-07
Revert "Use inline phpcs rules instead of phpcs.xml"

austinpray:
Reverts roots/bedrock#455


swalkinshaw:
Fine by me 👍 I don't feel too strongly and this file is useful to have if people actually customize it

PR #463, opened 2019-08-14
ini_set() expects parameter 2 to be string, int given

rubas:
`Uncaught TypeError: ini_set() expects parameter 2 to be string, int given`

https://www.php.net/manual/en/function.ini-set.php

### DIFF
--- a/config/application.php
+++ b/config/application.php
@@ -99,7 +99,7 @@ Config::define('DISALLOW_FILE_MODS', true);
  */
 Config::define('WP_DEBUG_DISPLAY', false);
 Config::define('SCRIPT_DEBUG', false);
-ini_set('display_errors', 0);
+ini_set('display_errors', '0');
 
 /**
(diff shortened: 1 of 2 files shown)


swalkinshaw:
😱 how was this never caught before. Looks correct though 👍

austinpray:
Yeah looks fine, peepin rn

austinpray:
Just tested it locally and it only occurs when `decare(strict_types=1)` is set at the top of application.php. Confirmed this with regular php 7 docker image php ini:

[screencapture-bedrock-test-phpinfo-php-2019-08-14-14_41_20.pdf](https://github.com/roots/bedrock/files/3503018/screencapture-bedrock-test-phpinfo-php-2019-08-14-14_41_20.pdf)

Also 

Before: https://3v4l.org/BrERf
After: https://3v4l.org/XM9SK

But yeah might as well use the right types

PR #464, opened 2019-08-16
Remove phpcs.xml from .gitattributes

Log1x:
With #455 revealing that `phpcs.xml` is actually used by people leading to #461 – it seems like it would be ideal if `phpcs.xml` is pulled down when creating a project.


austinpray:
Thx boyo

Issue #469, opened 2019-09-27
Env file does not parse password with # properly


Griefchief:
[x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
 [x] This request isn't a duplicate of an existing issue
 [x] I've read the [docs](https://roots.io/documentation/) and followed them (if applicable)
 [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

## Description

.env files do not parse a password with # in it properly. This gets parsed as the start of a comment, cutting of the password.

## Steps to reproduce

1. Set mysql db password to have # in it
2. add password to DB_PASSWORD field in .env
DB_PASSWORD=SomePass#Comment
3. refresh wordpress website, wordpress wp-db will complain.
![image](https://user-images.githubusercontent.com/5757704/65732889-3cd43080-e0cc-11e9-8c32-8a19d2ab134d.png)

**Expected behavior:** [Parse password properly or more realistically warn with a comment above to either escape the # if possible, or to not use #. I could not escape it...]

**Actual behavior:** [can't connect to db, see image above]

**Reproduces how often:** [100%]

## Versions

You can get this information from referencing `CHANGELOG.md`. Also, please include the OS and what version of the OS you're running.

## Additional information

Any additional information, configuration or data that might be necessary to reproduce the issue.

swalkinshaw:
Does quoting the value work?

```
DB_PASSWORD="SomePass#Comment"
```

Griefchief:
It does work. 

This can be nasty to debug, took me 2hrs and changing wp-db.php to echo the pass... I'd suggest a warning comment in env file if you think it's applicable.

swalkinshaw:
Closed by https://github.com/roots/bedrock/pull/471

PR #471, opened 2019-10-13
Document and suggest quoting .env var values

alwaysblank:
Issue #469 points out that characters following a `#` in the value of an unquoted string in an `.env` file will be parsed as a comment, which is demonstrably undesirable. https://github.com/vlucas/phpdotenv/issues/354 confirms that this is the intended behavior of the library Bedrock uses and that quoting the string in question solves the problem, but the Bedrock README doesn't make any mention of this and the example `.env` uses quotes only on the WordPress keys, implying that other values may not need quoting. 

For this PR I've wrapped all values in the example `.env` with single quotes. I've only used single quotes to match what the Roots key generator uses and the quotes already present in `.env.example`. So far as I can tell, dotenv considers `'` and `"` the same for the purposes of quoting values.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,18 +1,18 @@
-DB_NAME=database_name
-DB_USER=database_user
-DB_PASSWORD=database_password
+DB_NAME='database_name'
+DB_USER='database_user'
+DB_PASSWORD='database_password'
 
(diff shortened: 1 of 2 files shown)


swalkinshaw:
Thanks ❤️

PR #480, opened 2019-11-19
Stop auto sorting the packages

### LINKED ISSUE #479: Stop Auto sorting the composer.json
* [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
* [x] This request isn't a duplicate of an existing request
* [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

## Summary

Recently, the "sort-packages" value was added to the composer.json.

## Motivation

This wreaks havoc when trying to use Bedrock as a clean upstream git repo, which is how we manage all of our wordpress installs. When merging from upstream, git is no longer able to reconcile what bedrock is adding/updating, vs what my template has, and what that individual website has. This is seamless, but only when packages are sorting in the json based on order of inclusion.

Example json ordering, using a top-to-bottom approach for clean git management.
- Bedrock deps
- Common deps we add to all projects
- Individual deps per project

## Additional context

Each time we merge from upstream now we need to explicitly ignore this line, and manually reconcile the changes to json, every time. Removing this parameter from the json allows the developer to enforce their own structure, rather than forcing ABC order that doesn't serve a functional purpose. 

I feel like this addition was purely a cosmetic change, and works as a regression when trying to use within the context of CI/CD. My hope here is to keep Bedrock in a state where it can be easily integrated and updated in an automated manner.

devkinetic:
Fixes #479

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,8 +44,7 @@
   },
   "config": {
     "optimize-autoloader": true,
-    "preferred-install": "dist",
-    "sort-packages": true
+    "preferred-install": "dist"
   },
   "minimum-stability": "dev",
   "prefer-stable": true,


swalkinshaw:
Thanks 👍

Issue #474, opened 2019-10-21
$_SERVER var exposes db_name/user/password and salts


kupoback:
* [X] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
* [X] This request isn't a duplicate of an existing issue
* [X] I've read the [docs](https://roots.io/documentation/) and followed them (if applicable)
* [X] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

## Description

When working on a site, I noticed that if I was to `error_log()` the `$_SERVER` variable that the following additional items would be available (name/user/password removed for bug report):
```
[DB_NAME] => {db_name}
[DB_USER] => {user}
[DB_PASSWORD] => {userspwd}
[WP_ENV] => production
[WP_HOME] => https://domain.test
[WP_SITEURL] => https://wp-boilerplate.test/wp
[AUTH_KEY] => generateme
[SECURE_AUTH_KEY] => generateme
[LOGGED_IN_KEY] => generateme
[NONCE_KEY] => generateme
[AUTH_SALT] => generateme
[SECURE_AUTH_SALT] => generateme
[LOGGED_IN_SALT] => generateme
[NONCE_SALT] => generateme
```

## Steps to reproduce

1. Create a local install, use any environment
2. simply `var_dump()` or `error_log()` the `$_SERVER` var
3. See results

**Expected behavior:** These sensitive data shouldn't be available via that variable

**Actual behavior:** These sensitive data shouldn't be able to output

**Reproduces how often:** 100%

## Versions

Bedrock Install: 1.12.8: 2019-09-05
macOS: 10.14.6
laravel/valet 2.3.3

## Additional information

Basic Sage 9 with Bedrocks install on local.

QWp6t:
Looks like it would be a pretty straightforward change if we care to do this.

```diff
 /**
  * Expose global env() function from oscarotero/env
  */
 Env::init();
 
 /**
  * Use Dotenv to set required environment variables and load .env file in root
  */
-$dotenv = Dotenv\Dotenv::create($root_dir);
+$dotenv = Dotenv\Dotenv::create($root_dir, null, new Dotenv\Environment\DotenvFactory([
+    new Dotenv\Environment\Adapter\PutenvAdapter(),
+]));
 if (file_exists($root_dir . '/.env')) {
     $dotenv->load();
     $dotenv->required(['WP_HOME', 'WP_SITEURL']);
[...]

PR #485, opened 2019-12-09
Upgrade vlucas/phpdotenv to ^4.0.1 & fix breaking change

kellymears:
Ref #484

`vlucas/phpdotenv` > 4.0.0 has a breaking change as the initialization method has been split up and the old method has been removed.

The new methods require you to specify if the env definitions are immutable or mutable. 

This PR addresses this incompatibility by implementing the new `Dotenv::createImmutable` method. I don't think Bedrock is doing any overloading, so this is probably the better practice.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
   "require": {
     "php": ">=7.1",
     "composer/installers": "^1.7",
-    "vlucas/phpdotenv": "^3.6.0",
+    "vlucas/phpdotenv": "^4.0.1",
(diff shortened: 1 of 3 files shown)


QWp6t:
@kellymears Do you want to add this too?: https://github.com/roots/bedrock/issues/474#issuecomment-544714647

I was gonna create a PR (per swalk's request), but it would conflict with this PR, so might as well combine them.

austinpray:
no combine pls: keep these pr topics separate

PR #499, opened 2020-03-16
Add `WP_DEBUG_LOG`

TangRufus:
See: https://github.com/roots/trellis/pull/1160

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -13,6 +13,7 @@ DB_PASSWORD='database_password'
 WP_ENV='development'
 WP_HOME='http://example.com'
 WP_SITEURL="${WP_HOME}/wp"
+WP_DEBUG_LOG=/path/to/debug.log
 
 # Generate your keys here: https://roots.io/salts.html
 AUTH_KEY='generateme'
--- a/config/application.php
+++ b/config/application.php
@@ -106,6 +106,7 @@ Config::define('DISALLOW_FILE_MODS', true);
  * Debugging Settings
  */
 Config::define('WP_DEBUG_DISPLAY', false);
+Config::define('WP_DEBUG_LOG', env('WP_DEBUG_LOG') ?? false);
 Config::define('SCRIPT_DEBUG', false);
 ini_set('display_errors', '0');
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### Head
+
+* Add `WP_DEBUG_LOG` ([#499](https://github.com/roots/bedrock/pull/499))
+
 ### 1.13.0: 2020-02-17
 
 * Update to WordPress 5.3.2 ([#489](https://github.com/roots/bedrock/pull/489))


TangRufus:
> Is app/uploads the only guaranteed writeable place?

I believe so.

Updated default to `$webroot_dir . Config::get('CONTENT_DIR') . '/uploads/debug.log'`.

config/application.php
```diff
@@ -106,6 +106,7 @@
  * Debugging Settings
  */
 Config::define('WP_DEBUG_DISPLAY', false);
+Config::define('WP_DEBUG_LOG', env('WP_DEBUG_LOG') ?? $webroot_dir . Config::get('CONTENT_DIR') . '/uploads/debug.log');
```
swalkinshaw:
```php
WP_CONTENT_DIR . '/uploads/debug.log'
```

TangRufus: `Config::apply` is not yet applied at this point.
`WP_CONTENT_DIR` is most likely undefined.

swalkinshaw: Even with `Config::get('WP_CONTENT_DIR')`? 🤔

TangRufus:
```php
Config::define('WP_DEBUG_LOG', env('WP_DEBUG_LOG') ?? "${WP_CONTENT_DIR}/uploads/debug.log"); 

Config::apply();
==> PHP Notice:  Undefined variable: WP_CONTENT_DIR in ......./config/application.php on line 129

echo WP_DEBUG_LOG;
==> /uploads/debug.log
```

austinpray: can we avoid setting a default here?

swalkinshaw: 🤦‍♂ read the PR comments 😛 of course we can avoid it. the question is if a default is better than none

austinpray: ~~Can we~~ I recommend avoiding setting a default here. This could cause unintended consequences on high-availability setups. In the case WP_DEBUG is set to true on purpose or on accident this is a fail-open behavior rather than fail-safe.

TangRufus: > Even with Config::get('WP_CONTENT_DIR')? 

I think `Config::get` is possible.

---

How about we settle for:

```php
Config::define('WP_DEBUG_LOG', env('WP_DEBUG_LOG') ?? false); 
```

Reason: https://github.com/WordPress/WordPress/blob/7004afe4f4bac1fd17a142051832bdf6be8e6fcf/wp-includes/default-constants.php#L87-L89

austinpray: Yup that LGTM. fail-safe default.

TangRufus: And, what do we do on development?

Seems difficult to get a path for when using valet

austinpray: In development the developer just does whatever works best for their development setup.

TangRufus:
Thank you both.
Sorry my brain not functioning yesterday.

.env.example
```diff
@@ -13,6 +13,7 @@ DB_PASSWORD='database_password'
 WP_ENV='development'
 WP_HOME='http://example.com'
 WP_SITEURL="${WP_HOME}/wp"
+WP_DEBUG_LOG=/path/to/debug.log
```
QWp6t: Default value should be `true` imo so that it goes with default WordPress behavior.

PR #505, opened 2020-04-10
fix(debug-log): WP_DEBUG_LOG should be `true` by default

QWp6t:
Setting WP_DEBUG_LOG to `true` will cause WP to automatically log debug output to <WP_CONTENT>/debug.log (so app/debug.log). This is a sensible default.

See: https://wordpress.org/support/article/debugging-in-wordpress/#wp_debug_log

### DIFF
--- a/config/environments/development.php
+++ b/config/environments/development.php
@@ -4,10 +4,12 @@
  */
 
 use Roots\WPConfig\Config;
+use function Env\env;
 
 Config::define('SAVEQUERIES', true);
 Config::define('WP_DEBUG', true);
 Config::define('WP_DEBUG_DISPLAY', true);
+Config::define('WP_DEBUG_LOG', env('WP_DEBUG_LOG') ?? true);
 Config::define('WP_DISABLE_FATAL_ERROR_HANDLER', true);
 Config::define('SCRIPT_DEBUG', true);
 Config::define('DISALLOW_INDEXING', true);
--- a/.env.example
+++ b/.env.example
@@ -6,14 +6,16 @@ DB_PASSWORD='database_password'
 # When using a DSN, you can remove the DB_NAME, DB_USER, DB_PASSWORD, and DB_HOST variables
 # DATABASE_URL='mysql://database_user:database_password@database_host:database_port/database_name'
 
-# Optional variables
+# Optional database variables
 # DB_HOST='localhost'
 # DB_PREFIX='wp_'
 
 WP_ENV='development'
 WP_HOME='http://example.com'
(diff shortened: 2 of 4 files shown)


swalkinshaw:
Actually.. this goes counter to what we decided in https://github.com/roots/bedrock/pull/499.

If the default in the config is `false` then this example `.env` should mirror that?

QWp6t:
Yes, that's even better. Or remove the line entirely (or comment it out), which will fallback to default value of `false`. Just let me know what you prefer.

My reasoning for setting it to `true` was that we currently have non-false default value, so if we're going to have it non-false, then the safest non-false value is `true` instead of hard-coding a file path.

swalkinshaw:
We should just remove it imo. This doesn't even seem like a setting that will vary between environments? Since it primarily a development only setting.

QWp6t:
In that case, perhaps we should move it to `environments/development.php`. :thinking:

swalkinshaw:
I'm good with that since the two other debug related settings are already in there.

@austinpray I know our goal is to minimize env specific settings but I think this makes sense?

austinpray:
> @austinpray I know our goal is to minimize env specific settings but I think this makes sense?

Yeah LGTM!

TangRufus:
Why we don't encourage setting `WP_DEBUG_LOG` to a path out of web root?
Non-trellis users might have `example.com/app/debug.log` exposed to the public.

QWp6t:
Latest commit changes a few things:

* **.env** has option for specifying `WP_DEBUG_LOG`.
* **application.php** sets `WP_DEBUG_LOG` to `false`
* **environments/development.php** sets `WP_DEBUG_LOG` to value from **.env** or `true` to enable debug output during development

The last one there is probably what needs the most consideration. Do we want it `true` by default in development environment?

config/environments/development.php
```diff
@@ -8,6 +8,7 @@
 Config::define('SAVEQUERIES', true);
 Config::define('WP_DEBUG', true);
 Config::define('WP_DEBUG_DISPLAY', true);
+Config::define('WP_DEBUG_LOG', env('WP_DEBUG_LOG') ?? true);
```
andronocean, line 11: FYI, `env()` is undefined now with the update to oscarotero/env v2. Need to add `use function Env\env;` for this to work.

swalkinshaw:
@QWp6t think we just lost track of this, want to update it?

QWp6t:
Will do. I think I just need to add that `use function` line

swalkinshaw (approved):
I think this is an improvement 👍

PR #508, opened 2020-04-21
Composer: Add `only` option to wpackagist.org repository

TangRufus:
See:
- https://github.com/outlandishideas/wpackagist/pull/311
- https://github.com/composer/composer/pull/8755/
- https://php.watch/articles/composer-2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### Head
+
+* Composer: Add `only` option to wpackagist.org repository ([#508](https://github.com/roots/bedrock/pull/508))
+
 ### 1.13.2: 2020-04-01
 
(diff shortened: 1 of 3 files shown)


swalkinshaw:
This option is backwards compatible I guess? Composer just ignores it for non-2.0 versions?

TangRufus:
> This option is backwards compatible I guess?

Correct. Composer v1 ignores the `only` option.

---

Not to confuses with [canonical repositories](https://github.com/composer/composer/blob/master/doc/articles/repository-priorities.md#canonical-repositories):

> If a packages exists in a higher priority repository, it will now be entirely ignored in lower priority repositories. See repository priorities for details.
>
> -- https://github.com/composer/composer/blob/master/UPGRADE-2.0.md

PR #510, opened 2020-04-24
fix(mu-plugins): Fix invalid array access notice on 7.4 (Fixes #509)

### LINKED ISSUE #509: is_array check on null in updateCache method in bedrock-autoloader.php not compatible with php 7.4
* [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
* [x] This request isn't a duplicate of an existing issue
* [x] I've read the [docs](https://roots.io/documentation/) and followed them (if applicable)
* [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

## Description

This issue happens when using php 7.4 (as roots/trellis 1.4.0 does). In /web/app/mu-plugins/bedrock-autoloader.php on line 176 in the updateCache method of the Autoloader class. There is a check if `$this->cache['plugins']` is an array by using `is_array`. However, `$this->cache['plugins']` can be null at that point. As of php 7.4 `is_array` will no longer return `null` when the given variable is `null`, but throws a notice for invalid array access.

I was able to fix it by first checking if `$this->cache` is an array, like this:
`$rebuild           = !(is_array($this->cache) && is_array($this->cache['plugins']));`

## Steps to reproduce

1. Use Bedrock in combination with php 7.4 (roots/trellis 1.4.0 for example).
2. Make sure you can see php notices.
2. Add a mu-plugin in the mu-plugin folder.
3. See the notices appear when opening the site.

**Expected behavior:**
No php exceptions or notices when using bedrock and php 7.4

**Actual behavior:**
A php notice when using bedrock and php 7.4

## Versions

Bedrock 1.13.2
Trellis 1.4.0
OSX 10.14.6

Log1x:


ethanclevenger91:
@Log1x When updates get made to these, do you mind bumping the version on the bedrock-autload mu-plugin? That way we can quickly grep if we're caught up or not on older projects.

Log1x:
> @Log1x When updates get made to these, do you mind bumping the version on the bedrock-autload mu-plugin? That way we can quickly grep if we're caught up or not on older projects.

Hadn't even thought about it since it's never been bumped before – Done.

Issue #515, opened 2020-04-28
env function collide with illuminate/support


marmorag:
* [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
* [X] This request isn't a duplicate of an existing issue
* [X] I've read the [docs](https://roots.io/documentation/) and followed them (if applicable)
* [X] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

## Description

`env` function is colliding with `illuminate/support`.

The function `env` is redefined in package `illuminate/support` which is a dependency of `ilab/ilab-media-tools`, the issue is that the `env` function from `illuminate/support` is defined first and cause:

```
Uncaught Error: Class 'Dotenv\Environment\Adapter\EnvConstAdapter' not found in /var/www/html/vendor/illuminate/support/Env.php:67
```

## Steps to reproduce

1. `composer create-project roots/bedrock bedrock`
2. `composer require ilab/ilab-media-tools`
3. you get the error

**Expected behavior:** No error

**Actual behavior:** Class not found exception

**Reproduces how often:** 100%

## Versions

At least `1.13.4` I did not test with other version 
Reproducer on Ubuntu 18.04, and Debian 10 for original issue

## Additional information

You can find a reproducer [on this repository](https://github.com/marmorag/bedrock-ilabmedia-reproducer) a "fix" is on branch `fix` and [there's the PR](https://github.com/roots/bedrock/pull/516)

TangRufus:
Fixed in #530

Thanks for reporting

PR #516, opened 2020-04-28
fix: env function collide

marmorag:
As mentionned in #515 the `env` function is colliding with the one defined in `illuminate/support`, so why not to use directly the `Env::get` function.

### DIFF
--- a/config/application.php
+++ b/config/application.php
@@ -25,11 +25,6 @@ $root_dir = dirname(__DIR__);
 $webroot_dir = $root_dir . '/web';
 
 /**
- * Expose global env() function from oscarotero/env
- */
-Env::init();
-
-/**
  * Use Dotenv to set required environment variables and load .env file in root
  */
 $dotenv = Dotenv\Dotenv::createImmutable($root_dir);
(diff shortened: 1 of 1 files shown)


swalkinshaw:
Yep 👍 this is a breaking change since people could be using `env` by relying on this behaviour in any other file.

TangRufus:
> we should resist polluting the global namespace whenever possible.

Agree

> this is a breaking change

Agree

QWp6t:
We also have [roots/support](https://github.com/roots/support/blob/master/helpers.php#L5-L46), which has a namespaced `env()` function. ¯\\\_(ツ)\_/¯

austinpray:
> We also have [roots/support](https://github.com/roots/support/blob/master/helpers.php#L5-L46), which has a namespaced `env()` function. ¯\\\_(ツ)\_/¯
> 
> 

People can add that to their composer.json if they want 👍

tristanbes:
Thanks @marmorag for this PR. 

Global namespace should be avoided at all cost since you really can't rely on which `function` from different packages will be executed as shows the bug report attached to this PR.

tristanbes:
Hello. Is there anything blocking this PR ? We patched all our bedrocks FYI but what's preventing this from getting merged ?

austinpray:
closed in https://github.com/roots/bedrock/pull/530

PR #519, opened 2020-05-16
Use Bedrock Autoloader package 

retlehs:
Ref #299

https://github.com/roots/bedrock-autoloader


aaemnnosttv:
Awesome to finally see this happen! ❤️ 

Opened a PR against the autoloader repo which still has a few things that only belong here rather than in the package 🙂

retlehs:
@aaemnnosttv merged and updated this PR, thank you!

web/app/mu-plugins/bedrock-autoloader.php
```diff
@@ -11,233 +11,6 @@
 
 namespace Roots\Bedrock;
 
-if (!is_blog_installed()) {
-    return;
+if (is_blog_installed() && class_exists(Autoloader::class)) {
```
austinpray, line 14: :+1:

austinpray:
@aaemnnosttv do you care about this? https://github.com/roots/bedrock-autoloader/issues/7

vinkla:
I'm so glad this has finally happened! Great work everybody 🙌 

I've explored integrating Bedrock's autoloader into WordPlate. If we load the plugins within the `pre_option_active_plugins` hook there is no need to keep this as a mu-plugin. Instead we could load it from the `wp-config.php` file.

```php
public function __construct()
{
    if (isset(self::$instance)) {
        return;
    }

    self::$instance = $this;

    require_once ABSPATH . 'wp-includes/plugin.php';

    add_filter('pre_option_active_plugins', function () {
        $this->relativePath = '/../' . basename(WPMU_PLUGIN_DIR);

        if (is_admin()) {
            add_filter('show_advanced_plugins', [$this, 'showInAdmin'], 0, 2);
        }

        $this->loadPlugins();
    });
}
```

What are your thoughts on this?

austinpray:
@vinkla Thank you for taking a look! Im interested in your idea, but note that it cannot live in the constructor as a side-effect. It makes testing difficult and side-effects in constructors is a strong anti-pattern. Ref https://github.com/roots/bedrock-autoloader/issues/4

Working right now so can't respond in full.

vinkla:
> It makes testing difficult and side-effects in constructors is a strong anti-pattern.

Yes, this has been a pain point for us in WordPlate as well. Testing has been hard. I'll continue digging.

PR #521, opened 2020-05-20
Move disallow indexing mu-plugin to package

retlehs:
https://github.com/roots/bedrock-disallow-indexing

Ref #369 
Ref #59


config/application.php
```diff
@@ -95,6 +95,7 @@
 /**
  * Custom Settings
  */
+Config::define('DISALLOW_INDEXING', WP_ENV !== 'production');
```
TangRufus, line 98:
```suggestion
Config::define('DISALLOW_INDEXING', true);
```

Then add `Config::define('DISALLOW_INDEXING', false);` to `config/environments/staging.php` and `config/environments/development.php`

---

Alternatively:

```suggestion
Config::define('DISALLOW_INDEXING', env('DISALLOW_INDEXING') ?? WP_ENV !== 'production');
```

swalkinshaw:
Updated this

Issue #522, opened 2020-05-20
WordPress Site Health reports 'Should be improved' on default Bedrock install


retlehs:
## Description

On a default Bedrock install, the Site Health tool reports 'Should be improved'

## Steps to reproduce

1. Install Bedrock
2. Visit `/wp/wp-admin/site-health.php`

## Expected behavior

The Site Health tool shouldn't consider some checks to be relevant on a Bedrock install

## Actual behavior

The Site Health tool complains about:

### 2 critical issues

* Your site is set to display errors to site visitors (on development)
* Background updates are not working as expected

### 1 recommended improvements

* You should remove inactive themes (see https://github.com/roots/wordpress/issues/3)

yangm97:
When I do a default Bedrock install, instead of
- You should remove inactive themes

I get
- Have a default theme available

which gets listed under "critical" and is equally annoying :D

Levdbas:
Is a MU plugin the way to go that hooks into `site_status_tests` action and disables two tests?

`function disable_status_tests($tests)
{
   unset($tests['direct']['theme_version']); // disable theme update, theme inactive and default theme checks.
   unset($tests['async']['background_updates']); // disabled the background updates check.
   return $tests;
}`

christopherdarling:
Sharing here incase it helps anyone, I have made a PR with a solution to the themes problem — https://github.com/roots/bedrock/pull/555

PeterSafwat:
@christopherdarling Your solution should be included in the next release, having a ton of old themes on the server is a waste of resources and not secure.

retlehs:
https://packagist.org/packages/roots/wordpress-no-content is now available to use, fyi!

props to @LeoColomb and @swalkinshaw for their work on the new composer wordpress packages via https://github.com/roots/wordpress-packager

PR #540, opened 2020-07-29
add MU plugin that disables two site health status tests 'conflicting' with bedrock

Levdbas:
fixes

The PR removes the two Site Health Status notifications highlighted in red in the screenshot.

- Auto updates test, mentioned by @retlehs in #522 since Bedrock discourages auto-update by design, we disable the `background_updates` check.
- Theme update check , mentioned by @yangm97 in #522 By disabling `theme_version` we disable the test that checks if there is a default theme available, the default 'default theme' is twentytwenty. Since most bedrock installs will probably come without this default theme, this triggers the advice.
As a side effect the actual theme update checks aren't done as well. So I am not sure if disabling this test is the way to go.

The code run by the `theme_version` check can be found [here](https://github.com/WordPress/WordPress/blob/ef382e6aafaf4b826e8f60fec4bb826c9612da3e/wp-admin/includes/class-wp-site-health.php#L477)

This is my first PR attempt for a Roots project. So let me know if I missed something or something can be improved.

### DIFF
--- a/web/app/mu-plugins/disable-status-tests.php
+++ b/web/app/mu-plugins/disable-status-tests.php
@@ -1,0 +1,30 @@
+<?php
+
+/**
+ * Plugin Name:  Disable status tests
+ * Plugin URI:   https://github.com/roots/bedrock/
+ * Description:  Disable two status tests that conflict with the way Bedrock is intended.
+ * Version:      1.0.0
+ * Author:       Roots
+ * Author URI:   https://roots.io/
+ * License:      MIT License
+ */
+
+namespace Roots\Bedrock;
+
+add_filter('site_status_tests', __NAMESPACE__ . '\\disable_status_tests');
+
+/**
+ * Disable WordPress Health Status tests
+ * 
+ * Disables tests that currently conflict with the way WordPress is intened.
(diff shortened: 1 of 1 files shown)


swalkinshaw:
As another option, could we instead change the label and categorization of them? This is still probably a good solution, just wondering if it's better to leave them (so people know about their existence) but call out why they are "disabled".

Levdbas:
I think this would've been nice, but unfortunately the `site_status_tests` hook only lets you add test or remove existing tests, see https://developer.wordpress.org/reference/hooks/site_status_tests/ for the documentation and some examples for this hook.

The only 'workaround' would be to remove these mentioned tests and add new ones specific for a Bedrock installation. Now that I am brainstorming about this idea: For example a test that checks the WP_ENV constant and warns if the site is not on a subdomain and still has the WP_ENV on staging.

There are possibilities but out of the scope of this PR. But might be a nice thing to consider for increasing the UX/DX.

retlehs:
If you want to make a Composer package out of this I'd be happy to add it to our docs, but this isn't something we will be adding in Bedrock at this time

Issue #538, opened 2020-07-28
Use the new constant WP_ENVIRONMENT_TYPE


ArnaudBan:
* [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
* [x] This request isn't a duplicate of an existing request
* [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

## Summary

[WordPress 5.5 introduce environment](https://make.wordpress.org/core/2020/07/24/new-wp_get_environment_type-function-in-wordpress-5-5/).I think it could be a good idea to use the new variable `WP_ENVIRONMENT_TYPE` instead of `WP_ENV`

## Motivation

User are still going to add the new `WP_ENVIRONMENT_TYPE` so we are going to have two constant that have the same purpose.

palmiak:
I'm having some doubts here. I mean - having those separate is a bit strange, don't you think?

Or maybe we should set it like this:
if no WP_ENVIRONMENT_TYPE is set we should set WP_ENVIRONMENT_TYPE to be the same as WP_ENV. 

Or should they always be separated?

swalkinshaw:
Yes we'd likely support both transparently. So we'd set `WP_ENVIRONMENT_TYPE` according to `WP_ENV` and vice versa.

However, I think there's still hope that the `wp_get_environment_type` function will check `WP_ENV` as well.

palmiak:
Looking at this https://core.trac.wordpress.org/ticket/33161 - I don't think there is hope for WP_ENV :/

But aside - so in short:
- if both are defined - we just define theme as they are
- if only one is defined we define as the second one

I can do a PR for this.

palmiak:
Before I would push a PR - could you take a look:
```
Config::define('WP_ENVIRONMENT_TYPES', env('WP_ENVIRONMENT_TYPES') ? explode(',', env('WP_ENVIRONMENT_TYPES')) : ['production', 'staging', 'development']);

if (env('WP_ENV') || env('WP_ENVIRONMENT_TYPE')) {
    Config::define('WP_ENV', env('WP_ENV') ?: env('WP_ENVIRONMENT_TYPE'));
    Config::define('WP_ENVIRONMENT_TYPE', env('WP_ENVIRONMENT_TYPE') ?: env('WP_ENV'));
    define('WP_ENV', env('WP_ENV') ?: env('WP_ENVIRONMENT_TYPE'));
[...]

PR #543, opened 2020-08-13
Adds support for WP_ENVIRONMENT_TYPE

palmiak:
Based on #538 I added the support for new constants introduced in WP 5.5:
- `WP_ENVIRONMENT_TYPES`
- `WP_ENVIRONMENT_TYPE`

if `WP_ENV` is set but no `WP_ENVIRONMENT_TYPE` than `WP_ENVIRONMENT_TYPE = WP_ENV`
if `WP_ENVIRONMENT_TYPE` is set but no `WP_EVN` than `WP_ENV = WP_ENVIRONMENT_TYPE`

if none is set than `WP_ENV` is set to "production" and `wp_get_environment_type()` will return `production`.

### DIFF
--- a/config/application.php
+++ b/config/application.php
@@ -36,12 +36,26 @@ if (file_exists($root_dir . '/.env')) {
         $dotenv->required(['DB_NAME', 'DB_USER', 'DB_PASSWORD']);
     }
 }
+/**
+ * Set up Environment Types
+ * Default: production, staging, development
+ */
+Config::define('WP_ENVIRONMENT_TYPES', env('WP_ENVIRONMENT_TYPES') ? explode(',', env('WP_ENVIRONMENT_TYPES')) : ['production', 'staging', 'development']);
 
 /**
- * Set up our global environment constant and load its config first
- * Default: production
+ * Set up WP_ENV and WP_ENVIRONMENT_TYPE based on which
+ * constants are set
  */
-define('WP_ENV', env('WP_ENV') ?: 'production');
+if (env('WP_ENV') || env('WP_ENVIRONMENT_TYPE')) {
+    Config::define('WP_ENV', env('WP_ENV') ?: env('WP_ENVIRONMENT_TYPE'));
+    Config::define('WP_ENVIRONMENT_TYPE', env('WP_ENVIRONMENT_TYPE') ?: env('WP_ENV'));
+    define('WP_ENV', env('WP_ENV') ?: env('WP_ENVIRONMENT_TYPE'));
+} else {
+    Config::define('WP_ENV', 'production');
+    define('WP_ENV', 'production');
+}
+
+Config::define('DB_NAME', env('DB_NAME'));
 
 /**
  * URLs
--- a/config/application.php
+++ b/config/application.php
@@ -55,8 +55,6 @@ if (env('WP_ENV') || env('WP_ENVIRONMENT_TYPE')) {
     define('WP_ENV', 'production');
 }
 
-Config::define('DB_NAME', env('DB_NAME'));
-
 /**
  * URLs
  */


config/application.php
```diff
@@ -36,12 +36,26 @@
...
+if (env('WP_ENV') || env('WP_ENVIRONMENT_TYPE')) {
+    Config::define('WP_ENV', env('WP_ENV') ?: env('WP_ENVIRONMENT_TYPE'));
+    Config::define('WP_ENVIRONMENT_TYPE', env('WP_ENVIRONMENT_TYPE') ?: env('WP_ENV'));
+    define('WP_ENV', env('WP_ENV') ?: env('WP_ENVIRONMENT_TYPE'));
+} else {
+    Config::define('WP_ENV', 'production');
+    define('WP_ENV', 'production');
+}
+
+Config::define('DB_NAME', env('DB_NAME'));
```
LeoColomb, line 58: Is this last line intended? :wink:

palmiak: Thanks for spotting this :D Fixed.

config/application.php
```diff
@@ -36,12 +36,24 @@
         $dotenv->required(['DB_NAME', 'DB_USER', 'DB_PASSWORD']);
     }
 }
+/**
+ * Set up Environment Types
+ * Default: production, staging, development
+ */
+Config::define('WP_ENVIRONMENT_TYPES', env('WP_ENVIRONMENT_TYPES') ? explode(',', env('WP_ENVIRONMENT_TYPES')) : ['production', 'staging', 'development']);
```
ouun, line 43: As of WP 5.5.1 `WP_ENVIRONMENT_TYPES` [is deprecated](https://make.wordpress.org/core/2020/08/27/wordpress-environment-types/). Very short life...

The following environment types will be supported as of 5.5.1:

```
-     production – this is the default. A site that is running live, connected to the internet and reachable on the internet.
-     staging – this is what you would use for staging environments, probably both connected to and reachable on the internet.
-     development – this is what you would use for development environments that are reachable on the internet, we automatically enable WP_DEBUG on environments where this is the environment type.
-     local – added in 5.5.1, this (usually development) environment can reach the internet but is not be reachable from the internet.
```

As these include Bedrock defaults ['production', 'staging', 'development'], setting `WP_ENVIRONMENT_TYPES` is obsulete. 

```suggestion
```

ouun (changes requested):
As of WP 5.5.1 WP_ENVIRONMENT_TYPES [is deprecated](https://make.wordpress.org/core/2020/08/27/wordpress-environment-types/).

The following environment types will be supported as of 5.5.1:

```
-     production – this is the default. A site that is running live, connected to the internet and reachable on the internet.
-     staging – this is what you would use for staging environments, probably both connected to and reachable on the internet.
-     development – this is what you would use for development environments that are reachable on the internet, we automatically enable WP_DEBUG on environments where this is the environment type.
-     local – added in 5.5.1, this (usually development) environment can reach the internet but is not be reachable from the internet.

```

As these include Bedrock defaults ['production', 'staging', 'development'], setting WP_ENVIRONMENT_TYPES is obsulete.

swalkinshaw:
@palmiak are you interested in updating this to remove `WP_ENVIRONMENT_TYPES`?

TangRufus:
Should we throw errors when `WP_ENV` is not one of WordPress supported environment types, i.e: `production`, `staging`, `development` and `local`?

swalkinshaw:
WP just silently ignores unsupported ones and falls back to `production` ([ref](https://github.com/WordPress/wordpress-develop/blob/450df5728989a8f4412e8548f65e47691aaba3be/src/wp-includes/load.php#L234-L237)). If Bedrock wants to integrate with `WP_ENVIRONMENT_TYPE` then we'd have no choice but to only offer the same environments as WP core.

Otherwise, you'd end up with one `WP_ENV` and a different  `WP_ENVIRONMENT_TYPE` (the `production` default).

So if we're not okay with being limited to those env names, then we shouldn't integrate at all.

yangm97:
> WP just silently ignores unsupported ones and falls back to `production`

Same thing we do by setting production settings at `config/application.php` and diverging into environment specific files.

> So if we're not okay with being limited to those env names, then we shouldn't integrate at all.

Looking at the code it seems like the user can still set both variables independently if they so desire. 

IMO this PR can be merged, with @ouun's suggestion added.

swalkinshaw:
@yangm97 I'm not sure that's really analogous here?

With Bedrock you're free to create any new environment (such as "staging-2") with its own config and set `WP_ENV` to `staging-2` and everything works fine. But if we integrated with ` WP_ENVIRONMENT_TYPE`, then it would report `production` as the fallback since `staging-2` isn't supported.

> Looking at the code it seems like the user can still set both variables independently if they so desire.

Yes they can be set and used independently, but it could be confusing/misleading.

yangm97:
I see the problem now. What do you think about the way GitLab has solved this issue?

> By default, GitLab assumes a tier based on the environment name. Instead, you can use the deployment_tier keyword to specify a tier.

`WP_ENVIRONMENT_TYPE` is akin to Deployment Tier while `WP_ENV` is much like the "Environment Name". 

So theoretically you should be able to determine the `WP_ENVIRONMENT_TYPE` from `WP_ENV` but not the other way around.

- https://docs.gitlab.com/ee/ci/environments/#deployment-tier-of-environments
- https://docs.gitlab.com/ee/ci/yaml/index.html#environmentdeployment_tier

retlehs:
https://github.com/roots/bedrock/pull/668

PR #547, opened 2020-10-19
export-ignore github workflows

freezy-sk:
`.github` folder should not be in created projects


swalkinshaw:
Nice catch, thanks 👍

PR #555, opened 2020-10-30
ENH remove old wordpress themes

christopherdarling:
Not ideal, but a temporary solution to https://github.com/roots/wordpress/issues/3

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,18 @@
   },
   "scripts": {
     "post-root-package-install": [
-      "php -r \"copy('.env.example', '.env');\""
+      "php -r \"copy('.env.example', '.env');\"",
+      "composer run remove-old-wp-themes"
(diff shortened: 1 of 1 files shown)


BufferOverNo (approved):

mnightingale:
Wouldn't it make more sense in pre or post-autoload-dump?
I tested this in a bash shell you could remove all directories with `rm -rf -- web/wp/wp-content/themes/*/`

My concern would be cross compatibility (OS or possibly shell), hence why the existing post-root-package-install uses php to copy the .env.example file.

retlehs:
thank you for the PR, but we won't be recommending this approach officially 

this is currently blocked upstream on our end over at https://github.com/roots/wordpress/issues/3, and will update bedrock once that work has been completed

PR #563, opened 2020-11-30
Upgrade to vlucas/phpdotenv:5.2

dsturm:
This PR upgrades `vlucas/phpdotenv` to the current version of `5.2`. As in https://github.com/roots/acorn/pull/74#issuecomment-734279706 mentioned, Laravel 8.x and therefor `roots/acorn` with support for Laravel 8.x will need an upgrade of the package and [minor migration changes](https://github.com/vlucas/phpdotenv/blob/master/UPGRADING.md).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
   "require": {
     "php": ">=7.1",
     "composer/installers": "^1.8",
-    "vlucas/phpdotenv": "^4.1.8||^5.2",
+    "vlucas/phpdotenv": "^5.2",
     "oscarotero/env": "^2.1",
     "roots/bedrock-autoloader": "^1.0",
     "roots/wordpress": "5.5.3",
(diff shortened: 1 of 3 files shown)


TangRufus:
Question: Any reason to support v4?
If not for any important dependency conflicts, we can ditch the `Dotenv\Dotenv::createImmutable` condition.

dsturm:
> Question: Any reason to support v4?
> If not for any important dependency conflicts, we can ditch the `Dotenv\Dotenv::createImmutable` condition.

To be honest, I don't know if we really need v4. Since v5 only increase the PHP version to 7.1 and above I would say, we could ditch v4.

TangRufus:
>  we could ditch v4.

Please.

retlehs (approved):
thank you!

trainoasis:
How do I manually update to phpdotenv v5 from v3 without destroying everything? :D need it sorely for composer package requirements 🗡️

PR #572, opened 2021-02-03
Add WP_POST_REVISIONS as an option

mcnamee:
We more often than not, like to limit the number of post revisions in our various environments, in order to contain the size that the database can grow to (one client's wp_posts table is currently up to 200MB, after only 3 months of UAT).

Providing the ability to define this simply, via the `.env` will make the process of defining this per environment, much more convenient.

By default, it remains as `true` so not to effect the default behaviour.


mcnamee:
Hi there @retlehs - any chance you'd have a quick moment to review/merge this one in?

PR #577, opened 2021-03-05
feat(composer): improve repository syntax

SimJoSt:
Uses the same syntax that will be used when additional repositories are added, which converts the repository array to an object to allow for association.
This makes it easier to upgrade projects to new versions of the boilerplate down to road.


SimJoSt:
Retracted, as multiple repositories can be used with the array syntax. Key/value features of JSON objects don't have to be used.  
They can also cause problems with consistency, because they are considered unordered: https://getcomposer.org/doc/04-schema.md#repositories

PR #579, opened 2021-03-12
Remove Legacy WP Themes

alexstewartja:
Post-Install removal of themes considered unsupported by WP Core (not included in the latest [release](https://wordpress.org/releases/)).

An optimized version of  @christopherdarling's patch #555, temporarily solving roots/wordpress#3 in only one(1) line, instead of eight(8).

After all, 8 bits = 1 byte 😁

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,11 @@
   },
   "scripts": {
     "post-root-package-install": [
-      "php -r \"copy('.env.example', '.env');\""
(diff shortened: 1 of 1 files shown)


retlehs:
thank you for the PR, but we won't be recommending this approach officially 

this is currently blocked upstream on our end over at https://github.com/roots/wordpress/issues/3, and will update bedrock once that work has been completed

PR #586, opened 2021-03-30
Look for a .env.local for specific overrides.

trsteel88:
This is a different approach to override ENV variables (one that the Symfony framework follows).

Rather than executing a script from composer.json, .env becomes the default (and is committed).

Then, if you want to override specific variables, you would create a .env.local on the relevant environment. The .env.local is not committed.

This also solves an issue where ENV variables don't exist. e.g. if a dev creates a new ENV variable in .env.example. This isn't automatically appended to any other developers .env variables.

If they were to commit .env with a suitable default, then any other developers using the codebase will automatically use the suitable default defined.

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,26 @@
+DB_NAME='database_name'
+DB_USER='database_user'
+DB_PASSWORD='database_password'
+
+# Optionally, you can use a data source name (DSN)
+# When using a DSN, you can remove the DB_NAME, DB_USER, DB_PASSWORD, and DB_HOST variables
+# DATABASE_URL='mysql://database_user:database_password@database_host:database_port/database_name'
+
+# Optional variables
+# DB_HOST='localhost'
+# DB_PREFIX='wp_'
+
+WP_ENV='development'
+WP_HOME='http://example.com'
+WP_SITEURL="${WP_HOME}/wp"
+WP_DEBUG_LOG=/path/to/debug.log
+
+# Generate your keys here: https://roots.io/salts.html
+AUTH_KEY='generateme'
+SECURE_AUTH_KEY='generateme'
+LOGGED_IN_KEY='generateme'
+NONCE_KEY='generateme'
+AUTH_SALT='generateme'
+SECURE_AUTH_SALT='generateme'
+LOGGED_IN_SALT='generateme'
+NONCE_SALT='generateme'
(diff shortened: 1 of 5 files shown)


swalkinshaw:
🤔 I like parts of this idea.

Though overall I find it a bit weird to have a `.env` file committed with useless values. They aren't useful defaults, they're placeholders that *have* to be replaced. Usually the config model of having a `.local` override is for actual overrides.

@austinpray what do you think

JulienMelissas:
👍 to having a local override option, 👎  to committing the `.env`... that seems like a violation of best practices IMO.

MWDelaney:
`.env` is a build artifact. We shouldn't commit it.

trsteel88:
I wouldn't say they are useless defaults. The idea would be that you would put useful defaults in there and then if something needed to be reconfigured you would add it to the .env.local

I find it strange that .env.example is automatically copied to .env during a composer install.

If a developer creates a new ENV var then anyone who already has the project setup won't automatically get these variables which could then cause an exception in the application.

TangRufus:
I vote for following `vlucas/phpdotenv`'s expected [usage](https://github.com/vlucas/phpdotenv#usage-notes) plus allowing `.env.local` to override Trellis generated `.env`, i.e:
- keep `.env` git ignored
- keep `.env.example`
- remove https://github.com/roots/bedrock/blob/3dbfaec8b5c42d88cf401b80c291bbbad10991d9/composer.json#L62-L63
- load `.env.local` if exist
  ```diff
  - $dotenv = Dotenv\Dotenv::createUnsafeImmutable($root_dir);
  + $dotenv = Dotenv\Dotenv::createUnsafeImmutable($root_dir, ['.env', '.env.local'], false);
  ```


Use case:
Overriding DB credentials when using valet, local DB and `$ trellis dotenv`. So that my teammates are not required to follow my [db_password](https://github.com/roots/trellis/blob/17430191bb7211545eb63ba3ba989ee95c262c5f/group_vars/development/vault.yml#L10).

trsteel88:
Here is another solution: https://github.com/roots/bedrock/pull/587

Don't commit .env and just load .env.example first. If .env exists it will override project defaults.

swalkinshaw:
@trsteel88 thanks for proposing this and getting us thinking. https://github.com/roots/bedrock/pull/595 and https://github.com/roots/bedrock/pull/594 are the changes that will come out of all this discussion

PR #587, opened 2021-04-09
merge .env and .env.example

trsteel88:
This is an alternative to #586 

This PR removes the requirement to duplicate .env.example to .env on composer install.

When the application is loaded, it will load .env.example followed by .env.

.env will take preference and override any defaults. This is useful if a developer wanted to include any suitable defaults in their config.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -59,9 +59,6 @@
     "wordpress-install-dir": "web/wp"
   },
   "scripts": {
-    "post-root-package-install": [
-      "php -r \"copy('.env.example', '.env');\""
-    ],
     "test": [
       "phpcs"
     ]
--- a/composer.lock
+++ b/composer.lock
@@ -134,6 +134,10 @@
                 "zend",
                 "zikula"
             ],
+            "support": {
+                "issues": "https://github.com/composer/installers/issues",
+                "source": "https://github.com/composer/installers/tree/v1.10.0"
+            },
             "funding": [
                 {
                     "url": "https://packagist.com",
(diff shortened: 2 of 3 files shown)


swalkinshaw:
I like it better than the first one 👍 and again, I do like the idea of removing the composer copy script.

Honestly just unsure if this change is beneficial enough to outweigh the cost (every change has a cost, even improvements). I'll see what some others things, but regardless thanks for spending the time trying to improve Bedrock 🎉

QWp6t:
If this is the route we're gonna go, I think `.env.example` should be renamed to `.env.default`. @swalkinshaw thoughts?

trsteel88:
In my other PR I suggested having .env as the default and then if someone wants to override, they create a .env.local.

.env is commited
.env.local is created for overriding

This is common across many platforms. Here is some reading:

- https://symfony.com/doc/current/configuration.html
- https://medium.com/@tacomanator/environments-with-create-react-app-7b645312c09d
- https://github.com/entropitor/dotenv-cli/issues/37
- https://nextjs.org/docs/basic-features/environment-variables

swalkinshaw:
@trsteel88 thanks for proposing this and getting us thinking. https://github.com/roots/bedrock/pull/595 and https://github.com/roots/bedrock/pull/594 are the changes that will come out of all this discussion

trsteel88:
No worries @swalkinshaw. One other thing I have on my list is the caching of env vars. 

I believe at the moment .env is read on every request. It would be more efficient to create a php file in a cache on first load. php files will be cached by opcache which will improve performance.

PR #590, opened 2021-04-21
Sanitize WP_HOME

Androlax2:
Ensure that WP_HOME doesn't have unwanted whitespaces (before and after). And ensure there is no trailing slashes.

Had a problem on a website that came from here (leading whitespaces), took me 1 hour to find out where was the problem.

### DIFF
--- a/config/application.php
+++ b/config/application.php
@@ -46,7 +46,7 @@ define('WP_ENV', env('WP_ENV') ?: 'production');
 /**
  * URLs
  */
-Config::define('WP_HOME', env('WP_HOME'));
+Config::define('WP_HOME', rtrim(trim(env('WP_HOME')), '/'));
 Config::define('WP_SITEURL', env('WP_SITEURL'));
 
 /**


swalkinshaw:
This seemed completely reasonable at first thought... but it obscures a technically invalid config value which should be fixed instead of dynamically (and silently) sanitized. The `.env` file was partly chosen because it's portable and can be consumed by other tools as well, which may not do this sanitizing.

I'm sympathetic to the frustration you had with this, but I'm inclined to leave it as is since the issue hasn't come up before. If you or anyone else feels strongly about this, we can reconsider.

Androlax2:
> o

I understand, in this case. Wouldn't it be a better idea to blow an exception when the value is incorrect?

The problem was that I found myself in my site with %20 in front of the paths for example, and I did not understand at all where it came from I had to go back and debug a little bit each piece of code to finally go back to the config file.

Androlax2:
@swalkinshaw I understand, in this case. Wouldn't it be a better idea to blow an exception when the value is incorrect?

The problem was that I found myself in my site with %20 in front of the paths for example, and I did not understand at all where it came from I had to go back and debug a little bit each piece of code to finally go back to the config file.

swalkinshaw:
Throwing an exception would be better, yeah. Though technically wouldn't we want to validate a lot of other env vars as well?

Androlax2:
The other environment variables usually have more precise exceptions and therefore we know very quickly where the problem comes from.

If the database data is wrong => Error establishing a database connection on the frontend for example.

Whereas if you make a mistake on WP_HOME, it will break all the URLs of the site without really knowing what they are.

JuanBernal98:
Might consider using esc_url() to solve the broken links

PR #595, opened 2021-05-28
Remove post install .env copy

swalkinshaw:
`.env.example` is just an example and doesn't contain useful defaults. Copying this file can be confusing since it needs to be edited regardless. By not copying this file, we'll surface errors of missing config values sooner since they'll be no default values.

This is an alternative to https://github.com/roots/bedrock/pull/587 and https://github.com/roots/bedrock/pull/586

Docs PR: https://github.com/roots/docs/pull/476

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -65,9 +65,6 @@
     "wordpress-install-dir": "web/wp"
   },
   "scripts": {
-    "post-root-package-install": [
-      "php -r \"copy('.env.example', '.env');\""
-    ],
     "test": [
       "phpcs"
     ]


aaemnnosttv:
I find the `.env.example` to be useful as a guide (like the docs currently say) as to what values are needed and have actively added to it in the past as a reference for new devs.

Also, this change would break my [wp-cli `valet` command](https://github.com/aaemnnosttv/wp-cli-valet-command) which [replaces these defaults](https://github.com/aaemnnosttv/wp-cli-valet-command/blob/739ed285192d04086d2453c00ed7d9260fd609fa/src/Installer/BedrockInstaller.php#L38-L54) when creating a new Bedrock site. I'm not saying this change shouldn't happen for that reason only, but it would be nice to have some time to prepare an update before this is merged, if it goes through.

Jamiewarb:
I like this change. I think it's great for new players to Bedrock that might not be familiar with it. Following a README that asks to explicitly copy .env.example to .env will make them away of this file from the get-go.

README.md
```diff
@@ -71,7 +71,7 @@ Much of the philosophy behind Bedrock is inspired by the [Twelve-Factor App](htt
    ```sh
    $ composer create-project roots/bedrock
    ```
-2. Update environment variables in the `.env` file. Wrap values that may contain non-alphanumeric characters with quotes, or they may be incorrectly parsed.
+2. Create a `.env` config file (using `.env.example` as a guide) and set your config values. Wrap values that may contain non-alphanumeric characters with quotes, or they may be incorrectly parsed.
```
Jamiewarb, line 74: Might be we can follow suite with the first item in this list, and give the code for this as well, for quick copy/pasting.

E.G.:

2. Create a `.env` config file (using `.env.example` as a guide) and set your config values. Wrap values that may contain non-alphanumeric characters with quotes, or they may be incorrectly parsed.
   ```sh
   $ cp .env.example .env
   ```

swalkinshaw: Yeah I had considered that, I'll add it 👍

TangRufus:
Suggestion: if `.env` if not exists, `die()` with an error messge and link to roots.io/docs before https://github.com/roots/bedrock/blob/e9400200cdbdf00c97e7ad80802c48221ec7ac8d/config/application.php#L32 

Similar to https://github.com/roots/sage/blob/6d66debe17dc413cd9f276ddb42a368bf0dc844d/functions.php#L15

austinpray:
> Suggestion: if `.env` if not exists, `die()` with an error messge and link to roots.io/docs before
> 
> https://github.com/roots/bedrock/blob/e9400200cdbdf00c97e7ad80802c48221ec7ac8d/config/application.php#L32
> 
> Similar to https://github.com/roots/sage/blob/6d66debe17dc413cd9f276ddb42a368bf0dc844d/functions.php#L15

This doesn't work for dockerized environments where the environment variables are set by the container orchestrator (Kubernetes, heroku, etc). .env should only be used if you don't have first-class environment variable management.

TangRufus (approved):

ggiak:
@swalkinshaw 
check #674, should do the trick without hussle :)

PR #598, opened 2021-05-29
Don't leak env values into `$_SERVER`

### LINKED ISSUE #474: $_SERVER var exposes db_name/user/password and salts

swalkinshaw:
Updates to using a custom repository for `Dotenv` instead of the default which includes `ServerConstAdapter`.

The new custom repository *only* includes `EnvConstAdapter`.

The `$_SERVER` superglobal often gets dumped into logs or into monitoring services so it's better for security to avoid populating it with secrets contained in `.env`.

*Note*: this *could* be a breaking change for some users, but we at least need to ensure it's not breaking in the normal case.

Close #474

### DIFF
--- a/config/application.php
+++ b/config/application.php
@@ -39,8 +39,13 @@ if (file_exists($root_dir . '/.env')) {
         ? ['.env', '.env.local']
         : ['.env'];
 
-    $dotenv = Dotenv\Dotenv::createImmutable($root_dir, $env_files, false);
+    $repository = Dotenv\Repository\RepositoryBuilder::createWithNoAdapters()
+        ->addAdapter(Dotenv\Repository\Adapter\EnvConstAdapter::class)
+        ->addAdapter(Dotenv\Repository\Adapter\PutenvAdapter::class)
+        ->immutable()
+        ->make();
 
+    $dotenv = Dotenv\Dotenv::create($repository, $root_dir, $env_files, false);
     $dotenv->load();
 
     $dotenv->required(['WP_HOME', 'WP_SITEURL']);


config/application.php
```diff
@@ -29,7 +29,13 @@
  * Use Dotenv to set required environment variables and load .env file in root
  * .env.local will override .env if it exists
  */
-$dotenv = Dotenv\Dotenv::createUnsafeImmutable($root_dir, ['.env', '.env.local'], false);
+$repository = Dotenv\Repository\RepositoryBuilder::createWithNoAdapters()
+    ->addAdapter(Dotenv\Repository\Adapter\EnvConstAdapter::class)
+    ->addAdapter(Dotenv\Repository\Adapter\PutenvAdapter::class)
```
swalkinshaw, line 34: @QWp6t needed this as well to mirror the default behaviour and get everything working

montchr:
Is this still necessary to fix #474?

retlehs:
@montchr I think so! Thanks for the bump, I think this just got forgotten about 😅 will get this rebased and hopefully reviewed soon 

cc @QWp6t @tangrufus

Tofandel:
> Note: this could be a breaking change for some users, but we at least need to ensure it's not breaking in the normal case.

Given that this package doesn't have an upgrade path because it's a boilerplate, I wouldn't worry about breaking changes

PR #599, opened 2021-06-26
Update composer.json

saddam-azad:
This makes the installer fetch the latest version of Wordpress upon Bedrock installation.


QWp6t:
x-ref: #297

PR #602, opened 2021-07-30
Fixes #601 - Load .env.local only if the file exists

### LINKED ISSUE #601: .env.local failed to open stream php error when WP_ENV='development' and .env.local does not exist.
Fresh bedrock install. 

When you only have .env (not .env.local) and your WP_ENV is set to development, WordPress admin will have an extra space at the top of the screen. This is caused by an invisible failed to open stream error when .env.local is not found.

To verify the cause, you can simply go into your application.php file and remove '.env.local' from the array on line 32. This removes php error, and the space at the top of WordPress admin.

mrgesco:
If the file `.env.local` doesn't exists, getting a warning
`Warning: file_get_contents(/var/www/.env.local): failed to open stream: No such file or directory`

Added a check to load the `.env.local` only if the file exists

### DIFF
--- a/config/application.php
+++ b/config/application.php
@@ -29,8 +29,12 @@ $webroot_dir = $root_dir . '/web';
  * Use Dotenv to set required environment variables and load .env file in root
  * .env.local will override .env if it exists
  */
-$dotenv = Dotenv\Dotenv::createUnsafeImmutable($root_dir, ['.env', '.env.local'], false);
+$dotenv = Dotenv\Dotenv::createUnsafeImmutable($root_dir, ['.env'], false);
 if (file_exists($root_dir . '/.env')) {
+    if(file_exists($root_dir . '/.env.local')) {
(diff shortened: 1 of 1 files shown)


TangRufus:
For reference: https://github.com/roots/bedrock/commit/b2781b890e5f988f847e716e6abbc48e34c3cc5c#commitcomment-54196503

swalkinshaw:
🤔 does this actually do what we want? `.env.local` should act as an override only, but this would completely replace the loader for `.env`. Meaning if `.env.local` existed, `.env` wouldn't be loaded at all.

The solution proposed in https://github.com/roots/bedrock/commit/b2781b890e5f988f847e716e6abbc48e34c3cc5c#commitcomment-54196503 looks more correct

mrgesco:
@swalkinshaw your right this would replace the .env

PR #603, opened 2021-07-30
Fixes #601 - .env.local failed to open stream php warning

### LINKED ISSUE #601: .env.local failed to open stream php error when WP_ENV='development' and .env.local does not exist.

fab120:
For reference: [b2781b8#commitcomment-54196503](https://github.com/roots/bedrock/commit/b2781b890e5f988f847e716e6abbc48e34c3cc5c#commitcomment-54196503)

I also removed this `if`
https://github.com/roots/bedrock/blob/master/config/application.php#L33

If .env file is not present a `\Dotenv\Exception\InvalidPathException` will be raised.

Fixes #601

### DIFF
--- a/config/application.php
+++ b/config/application.php
@@ -29,7 +29,11 @@ $webroot_dir = $root_dir . '/web';
  * Use Dotenv to set required environment variables and load .env file in root
  * .env.local will override .env if it exists
  */
-$dotenv = Dotenv\Dotenv::createUnsafeImmutable($root_dir, ['.env', '.env.local'], false);
+$env_files = file_exists($root_dir . '/.env.local')
+    ? ['.env', '.env.local']
+    : ['.env'];
+
+$dotenv = Dotenv\Dotenv::createUnsafeImmutable($root_dir, $env_files, false);
 if (file_exists($root_dir . '/.env')) {
     $dotenv->load();
     $dotenv->required(['WP_HOME', 'WP_SITEURL']);


swalkinshaw:
My only question is if we really want to remove the file existence check. I *think* that was originally done to allow more flexibility with actual env vars. `.env` was chosen because it's portable and doesn't have to be used. You can inject env vars via your web server for example.

This PR actually forces people to have a `.env` file which could be a regression for some users.

TangRufus:
@austinpray mentioned there is no `.env` file on a docker setup.
Therefore, we should allow omitting `.env` files.

fab120:
> @austinpray mentioned there is no `.env` file on a docker setup.
> Therefore, we should allow omitting `.env` files.
> See: [#595 (comment)](https://github.com/roots/bedrock/pull/595#issuecomment-850844041)

Oh yes. Missed that.
Removed the second commit that was deleting the `if`.

swalkinshaw:
Thank you 👏  we'll get a release out asap

fab120:
> My only question is if we really want to remove the file existence check. I _think_ that was originally done to allow more flexibility with actual env vars. `.env` was chosen because it's portable and doesn't have to be used. You can inject env vars via your web server for example.
> 
> This PR actually forces people to have a `.env` file which could be a regression for some users.

Totally agree and honestly i didn't think at this use case when i was looking at that code.
A solution should use `safeLoad` instead of `load`: 
```php
$dotenv = Dotenv\Dotenv::createUnsafeImmutable($root_dir, $env_files, false);
$dotenv->safeLoad();
$dotenv->required(['WP_HOME', 'WP_SITEURL']);
if (!env('DATABASE_URL')) {
    $dotenv->required(['DB_NAME', 'DB_USER', 'DB_PASSWORD']);
}
```
Therefore, if the file is missing no error is raised.
The two `required` check will also pass because `Dotenv` is checking the environment and not the content of the `.env` file

swalkinshaw:
@fab120 that's nicer! Feel free to open a PR if you want since that's an improvement

PR #609, opened 2021-10-19
Update to default branch

huubl:


mrkaluzny:
@swalkinshaw Problem when installing wp-cli 
```
Root composer.json requires roave/security-advisories dev-latest, found roave/security-advisories[dev-latest] but the package is fixed to dev-master (lock file version) by a partial update and that version does not match. Make sure you list it as an argument for the update command.
```

retlehs:
@mrkaluzny what exact command did you run that resulted in that error?

mrkaluzny:
@retlehs 

```
composer require wp-cli/wp-cli-bundle
```

It seems they haven't updated the branch in [wp-cli composer.json](https://github.com/wp-cli/wp-cli/blob/master/composer.json)

retlehs:
@mrkaluzny i'm not able to reproduce. can you please open a topic over at https://discourse.roots.io/ if you can't resolve this problem?

mrkaluzny:
@retlehs Hmm it works with fresh install of bedrock every time on my end, I'll open it :)

mrkaluzny:
@retlehs I've pushed update to wp-cli repository it's working as expected now :)

PR #636, opened 2022-04-28
Require project autoloader in wp-cli.yml

ethanclevenger91:
If you're using the global WP-CLI, it won't include the project's Composer autoloader until WordPress itself loads and Bedrock's own code to include the autoloader happens. This means that if you require third-party WP-CLI commands in a project and attempt to run them via a global copy of WP-CLI, those commands will not load before WordPress, even if they should (via `@when before_wp_load`).

Use case:
- Have WP-CLI available globally via their install instructions
- Create a Bedrock project
- `composer require aaemnnosttv/wp-cli-dotenv-command`
- Run `wp help`
- See `dotenv` under available commands
- Run `wp dotenv init --template=.env.example --with-salts` (this command is configured to be available [before_wp_load](https://github.com/aaemnnosttv/wp-cli-dotenv-command/blob/master/src/WP_CLI/DotenvCommand.php#L32))

**Expected behavior:**
Create .env file from template

**Actual behavior:**
Error because database could not be reached

Requiring the local autoloader as part of WP-CLI's configuration makes commands available when they're supposed to be available, regardless of which instance of WP-CLI you're using.


ethanclevenger91:
Related: https://github.com/aaemnnosttv/wp-cli-dotenv-command/issues/12

QWp6t (approved):

PR #638, opened 2022-05-23
🔥 remove register theme directory mu-plugin

retlehs:
### to be merged after wordpress 6.0

this mu-plugin will no longer be required as of wordpress 6.0. the `roots/wordpress` composer package is switching to use the "no-content" distribution, and bedrock will then be requiring the twentytwentytwo theme as a dependency

read more at https://roots.io/updates-to-roots-wordpress-composer-package/


retlehs:
#639

PR #658, opened 2022-11-09
fix: can disable revisions by setting false or 0

TimeZHero:
Hi, noticed that the default setup for application.php does not allow to turn off revisions through the value of `WP_POST_REVISIONS`. Setting either false or 0 results on enabling the feature completely, which is pretty confusing.


swalkinshaw:
Thanks @TimeZHero

PR #665, opened 2023-01-02
🔧 Bump requirements to PHP 8.0

retlehs:
https://www.php.net/supported-versions.php
https://endoflife.date/php

I'd prefer to bump this to PHP 8.1, considering active support for PHP 8.0 ended on 2022-11-26, but WordPress & WP-CLI _still_ don't fully support it at this time

Ref **https://roots.io/wordpress-php-8-1-support-not-found/**
Ref https://core.trac.wordpress.org/ticket/54504 (coming in WordPress 6.2)
Ref https://github.com/wp-cli/wp-cli/issues/5623

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        php-version: ['7.4', '8.0', '8.1', '8.2']
+        php-version: ['8.0', '8.1', '8.2']
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


swalkinshaw (approved):

drzraf:
I believe it's an error.
Those relying on PHP 7.x (and whatever the reasons are, it's **their** problem) should be given the choice. If a dependency **actually** needs 8.x, it's fine but please don't arbitrarily increase the version number. WP still supports a broad range of OHO versions (including 7.x) and that's fine.

retlehs:
It’s actually not fine that WordPress supports versions of PHP that aren’t actively supported.

Bedrock is just a boilerplate. Feel free to change the value to whatever you’d like for your own projects — no one is forcing you to use a certain version of PHP. We’ll be bumping this to PHP 8.1 now that WordPress 6.2 is out.

PR #668, opened 2023-02-15
Support WP_ENVIRONMENT_TYPE

voronkovich:
This PR implements proposal suggested by @swalkinshaw: https://github.com/roots/bedrock/pull/627#issuecomment-1329295366

### DIFF
--- a/config/application.php
+++ b/config/application.php
@@ -49,6 +49,13 @@ if (file_exists($root_dir . '/.env')) {
 define('WP_ENV', env('WP_ENV') ?: 'production');
 
 /**
+ * Infer WP_ENVIRONMENT_TYPE based on WP_ENV
+ */
+if (!env('WP_ENVIRONMENT_TYPE') && in_array(WP_ENV, ['production', 'staging', 'development'])) {
+    Config::define('WP_ENVIRONMENT_TYPE', WP_ENV);
+}
+
+/**
  * URLs
  */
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -84,6 +84,7 @@ Much of the philosophy behind Bedrock is inspired by the [Twelve-Factor App](htt
   - `DB_HOST` - Database host
   - Optionally, you can define `DATABASE_URL` for using a DSN instead of using the variables above (e.g. `mysql://user:password@127.0.0.1:3306/db_name`)
 - `WP_ENV` - Set to environment (`development`, `staging`, `production`)
+- `WP_ENVIRONMENT_TYPE` - Set to [environment type](https://developer.wordpress.org/reference/functions/wp_get_environment_type/)
```
swalkinshaw, line 87: Let's omit this for now since by default it will be set automatically (and in ~98% of cases?). Don't want people thinking they need to set it.

voronkovich: @swalkinshaw , ok. I've removed that line from README.

swalkinshaw:
Thanks @voronkovich

yangm97:
@swalkinshaw why this PR got merged instead of #543?

swalkinshaw:
That PR was outdated using the deprecated `WP_ENVIRONMENT_TYPES` (plural). It also set `WP_ENV` based on `WP_ENVIRONMENT_TYPE` which this only does the opposite so it's simpler.

PR #670, opened 2023-02-28
Don't instantiate Dotenv if .env file not exists

voronkovich:
In `production` environment `.env` files are never used, so we don't need to create `Dotenv` instance if `.env` file not exists.

### DIFF
--- a/config/application.php
+++ b/config/application.php
@@ -29,13 +29,15 @@ $webroot_dir = $root_dir . '/web';
  * Use Dotenv to set required environment variables and load .env file in root
  * .env.local will override .env if it exists
  */
-$env_files = file_exists($root_dir . '/.env.local')
-    ? ['.env', '.env.local']
-    : ['.env'];
-
-$dotenv = Dotenv\Dotenv::createUnsafeImmutable($root_dir, $env_files, false);
 if (file_exists($root_dir . '/.env')) {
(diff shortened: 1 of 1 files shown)


swalkinshaw:
This optimization makes sense but I think it changes the behaviour when a `.env` doesn't exist.

Before: dotenv would throw an exception letting you know it's required
After: no exception because this is only within a `file_exists` conditional

config/application.php
```diff
@@ -29,13 +29,15 @@
...
-
-$dotenv = Dotenv\Dotenv::createUnsafeImmutable($root_dir, $env_files, false);
 if (file_exists($root_dir . '/.env')) {
+    $env_files = file_exists($root_dir . '/.env.local')
+        ? ['.env', '.env.local']
+        : ['.env'];
+
+    $dotenv = Dotenv\Dotenv::createUnsafeImmutable($root_dir, $env_files, false);
+
     $dotenv->load();
```
voronkovich, line 39: @swalkinshaw, I think you're wrong. The exception could be thrown here when `$dotenv->load()` is calling. But it is already inside `if` statement. BTW, anyway I think that `.env` file should not be required.

swalkinshaw: Ah yes, that's correct 👍

swalkinshaw:
Thanks @voronkovich

PR #674, opened 2023-04-01
chore: creates .env only if not exists

ggiak:
Following up to #595 

Given post install creates an .env based on .env.example 
it is useful to happen only if .env doesn't exist 
similar to what other open source projects do.


swalkinshaw:
Thanks @ggiak but I'm going to go forward with https://github.com/roots/bedrock/pull/595

Issue #679, opened 2023-06-12
ADD: Local environment in application.php


DSGND:
### Summary

CURRENTLY: 
- It misses the `local` environment here: https://github.com/roots/bedrock/blob/052bf9d60b08a5f00273f6846999709e900eed27/config/application.php#L56
- So when set ENV=local in .env, wp_get_environment_type() return `production`
- .env.local is not used

TO DO:
- According to [official documentation](https://developer.wordpress.org/reference/functions/wp_get_environment_type/), add 'local' in the array as so : 
```
if (!env('WP_ENVIRONMENT_TYPE') && in_array(WP_ENV, ['production', 'staging', 'development', 'local'])) {
    Config::define('WP_ENVIRONMENT_TYPE', WP_ENV);
}
```

DSGND:
The enhancement is merged : https://github.com/roots/bedrock/blob/9223c3b2236e9594a8136bfda93db4dfd7574717/config/application.php#L56

The issue can be closed.

PR #681, opened 2023-06-12
ADD: `local` environment in application.php

DSGND:
To fix this issue : https://github.com/roots/bedrock/issues/679

According to [official documentation](https://developer.wordpress.org/reference/functions/wp_get_environment_type/), add 'local' in the array in application.php.


swalkinshaw:
Thanks @DSGND

Issue #585, opened 2021-03-28
production.php missing?


strarsis:
* [x] I've read the [guidelines for Contributing to Roots Projects](https://github.com/roots/.github/blob/master/CONTRIBUTING.md)
* [x] This request isn't a duplicate of an existing issue
* [x] I've read the [docs](https://roots.io/documentation/) and followed them (if applicable)
* [x] This is not a personal support request that should be posted on the [Roots Discourse](https://discourse.roots.io/) community

## Description
The `production.php` is missing from `config/environments/` in the repository. It would be nice to have a basic configuration for disabling error display, file mods, etc.

austinpray:
As of https://github.com/roots/bedrock/pull/380, application.php is production by default. Then development, staging, whatever are environments where you make the minimum amount of deviations from production in order to make your life easier.

Before, if you somehow messed up setting WP_ENV to "production", the behavior was fail-open. Now it's fail-safe where it will default to production settings if you mess up.

PR #691, opened 2023-10-23
Add file to configure production environments

natanael-aguiar:
When I had my first contact with Bedrock I felt lost as to whether or not I should create a production file and define my production environment settings in it, I believe more people should have it too. I took courage and created a simple file for the production environment, which contains some basic settings that are up to the developer to edit and configure to their preference. This way we will have more agility when uploading our applications to a production environment. I hope this can help in some way.

Print attached

image

### DIFF
--- a/config/environments/production.php
+++ b/config/environments/production.php
@@ -1,0 +1,52 @@
+<?php
+
+/**
+ * Configuration overrides for WP_ENV === 'production'
+ */
+
+use Roots\WPConfig\Config;
+use function Env\env;
+
+/* Debug */
+Config::define('SAVEQUERIES', false);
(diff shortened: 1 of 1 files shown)


QWp6t:
Hey thanks for your willingness to contribute to Bedrock.

We intentionally [removed `environments/production.php`](https://github.com/roots/bedrock/pull/585).

This is because our `application.php` file is intended to be [production by default](https://github.com/roots/bedrock/issues/380).

So with that said, I don't think that this change is the right fit for Bedrock.

However, a couple of your changes might be worthy of discussion for adding to `application.php`. Feel free to open an issue if you want to discuss any potential changes.

Thanks again for being part of the community! 🙏

PR #709, opened 2024-02-02
fix: The `set-output` command is deprecated and will be disabled soon.

dr5hn:
This PR Fixes warnings in CI action.

Screenshot 2024-02-02 at 7 04 08 PM

Ref: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


swalkinshaw:
Thank you!

PR #717, opened 2024-04-24
🔧 Disable script concatenation by default

retlehs:
This PR sets `CONCATENATE_SCRIPTS` to false by default which prevents WordPress from concatenating scripts

Ref https://discourse.roots.io/t/upstream-timed-out-for-load-styles-php/26973

### DIFF
--- a/config/application.php
+++ b/config/application.php
@@ -124,6 +124,9 @@ Config::define('DISALLOW_FILE_MODS', true);
 // Limit the number of post revisions
 Config::define('WP_POST_REVISIONS', env('WP_POST_REVISIONS') ?? true);
 
+// Disable script concatenation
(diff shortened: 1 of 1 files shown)


swalkinshaw (approved):

tangrufus:
Should we take it one step further to disable `COMPRESS_SCRIPTS`, `COMPRESS_CSS` and `ENFORCE_GZIP` as well?

The CDN / web servers (e.g: nginx, apache, etc) should take care of those.

slackday:
I started getting problems with this after updating nginx from 1.21.7 to 1.27.2. Not with timeout to load-styles.php like many people report but nginx would just return connection reset by peer.

Setting CONCATENATE_SCRIPTS solved the problem. Not sure why it happens but seems WordPress does something not compatible with latest nginx.

PR #726, opened 2024-06-21
Avoid redefining `DISALLOW_FILE_EDIT`

joshuafredrickson:
Some security plugins will automatically edit `wp-config.php` and then throw an error:

```
Fatal error: Uncaught Roots\WPConfig\Exceptions\ConstantAlreadyDefinedException:
Aborted trying to redefine constant 'DISALLOW_FILE_EDIT'.
`define('DISALLOW_FILE_EDIT', ...)` has already been occurred elsewhere.
```

I know this happens with at least two plugins:

- Solid Security Pro (occurs automatically upon activation)
- Sucuri Security (occurs when option is activated in settings)

Of course, adding these constants to `wp-config.php` in Bedrock isn't ideal, but this check at least keeps the fatal error from triggering when working with these plugins. 

I'm not sold this is something that Bedrock needs to address... It's just something I've encountered a few times in the wild. I could be convinced either way.

### DIFF
--- a/config/application.php
+++ b/config/application.php
@@ -116,7 +116,9 @@ Config::define('AUTOMATIC_UPDATER_DISABLED', true);
 Config::define('DISABLE_WP_CRON', env('DISABLE_WP_CRON') ?: false);
 
 // Disable the plugin and theme file editor in the admin
-Config::define('DISALLOW_FILE_EDIT', true);
+if (!defined('DISALLOW_FILE_EDIT')) {
+    Config::define('DISALLOW_FILE_EDIT', true);
+}
 
 // Disable plugin and theme updates and installation from the admin
 Config::define('DISALLOW_FILE_MODS', true);


thedavidthomas:
+1 - Solid Security Pro (previously called iThemes Security Pro) user here. 
I have been doing this programatically immediately after creating a bedrock project since `DISALLOW_FILE_EDIT` is added automatically to `wp-config.php` on activation.

andronocean:
I don't think I like this idea. The point of throwing the exception is that you ought to know — _very quickly_ — if constants are getting defined in multiple places. `if (!defined('DISALLOW_FILE_EDIT'))` would also skip the definition if `define('DISALLOW_FILE_EDIT', false)` was added — in which case one's site is now wide open to admin editing without anyone knowing about it.

It seems to me the security plugins are "doing it wrong", and should be checking whether the constant is already defined before modifying wp-config.php.

PR #728, opened 2024-06-26
Display correct bedrock-autoloader version

chrillep:
https://github.com/roots/bedrock/blob/8df74d1dfdd60953db939ac7cbd3d134310aa863/composer.lock#L342-L343

Maybe there is a better way of fetching version from lockfile or adding it to roots/bedrock-autoloader per default. But until then...


tangrufus:
Or, we can just remove the `Version` line from plugin header.

chrillep:
> Or, we can just remove the `Version` line from plugin header.

i just happend to stumble upon this when i was looking for something else. :)

retlehs:
Good idea @tangrufus :+1:

@chrillep Are you not open to iterating on this PR after receiving feedback? If so, just close it.

chrillep:
> Good idea @tangrufus 👍
> 
> @chrillep Are you not open to iterating on this PR after receiving feedback? If so, just close it.

sorry went on vacation.

I removed the Version line from plugin header.

PR #734, opened 2024-08-22
Add editor configs to gitignore

NietThijmen:
Changing the default gitignore a little bit to remove some folders added by text editors/IDE's

This makes sure these are never pushed, which makes working in a team a lot harder as overwriting each others editor configs isn't preferable.


swalkinshaw:
These should be added to people's global gitignore files (`$HOME/.gitignore`) and instead of polluting every individual repo in my opinion.

PR #746, opened 2024-12-13
🎨 PSR-12 coding standards, switch from PHPCS to Pint

retlehs:


vinkla:
I would recommend switching to [PER-CS](https://www.php-fig.org/per/coding-style/) which is the successor to PSR-12.

retlehs:
Looks like we'll have to bump to a minimum of PHP 8.1 in order to grab a Pint release that supports that, which I'm fine with. Thanks for the call out @vinkla 🙏

vinkla:
Another idea is to use PHP-CS-Fixer directly. The setup is similar to Pint, which is a wrapper library. You won't even have to use a config file if you only want it for checking pull requests. 

https://github.com/vinkla/extended-acf/blob/master/.github/workflows/php-cs-fixer.yml

retlehs:
#749

PR #752, opened 2024-12-16
Added auth.json

tombroucke:
auth.json is used to authenticate requests to private composer repsitories and should never be committed to the repo.


retlehs (approved):
Thank you!

PR #757, opened 2025-01-24
Added support for WP-CLI with superuser privileges - WP_CLI_ALLOW_ROOT: 1

Daniel217D:
It is now possible to run commands in the app container without constantly specifying the --allow-root flag.


swalkinshaw:
Thank you!

PR #775, opened 2025-07-03
add Bedrock Composer Store plugin to manage Composer dependencies

CallMeLeon167:
# Add "Add to Composer" buttons to WordPress plugin store

## What's this about?

Ever been in the WordPress admin, seen a cool plugin, and thought "I want this in my Bedrock project"? Then you probably did one of these dances:

1. Manually find the plugin slug
2. Edit `composer.json` by hand or run `composer require wpackagist-plugin/plugin-name` (hope you got the slug right)
3. Run `composer install`

Well, no more! This adds "Add to Composer" buttons right next to the install buttons in the plugin store.

## What it does

- Adds a blue "Add to Composer" button next to every plugin install button 
- Fetches the current version from WordPress.org API  
- Adds entries like `"wpackagist-plugin/contact-form-7": "^5.7.7"` to your composer.json
- Shows "Already in Composer" for plugins you've already added

## Why should this be in Bedrock core?

**This solves a core Bedrock workflow problem.** Here's why it belongs in the main project:

### 1. **It's a fundamental Bedrock workflow issue**
Every Bedrock developer faces this. The WordPress admin and Composer workflow are disconnected, leading to:
- Accidental plugin installs that break deployments  
- Manual composer.json editing
- New team members installing plugins the "wrong way"

### 2. **It bridges the WordPress/Composer gap**
Bedrock's whole philosophy is "WordPress with modern tools." This plugin makes that seamless by bringing Composer functionality into the familiar WordPress admin interface.

### 3. **It prevents common mistakes**
Without this, developers (especially new ones) will inevitably install plugins through WordPress admin, causing deployment issues and environment inconsistencies.

### 4. **It's lightweight and focused**
- Single file MU plugin (~200 lines)
- No external dependencies
- Only loads on plugin-install.php
- Zero impact on site performance

## The bigger picture

Bedrock is about making WordPress development better. This plugin removes a major friction point that every Bedrock team encounters daily. It's not just a "nice to have" - it's addressing a fundamental workflow disconnect that exists in every Bedrock project.

Plus, it's the kind of feature that makes people go "Why didn't WordPress have this already?" 😄

image

### DIFF
--- a/web/app/mu-plugins/bedrock-composer-store.php
+++ b/web/app/mu-plugins/bedrock-composer-store.php
@@ -149,7 +149,7 @@ class BedrockComposerStore
                         success: function(r) {
                             if (r.success) {
                                 $btn.addClass('disabled').text(bedrock_ajax.added_text);
-                                $('<div class="notice notice-success is-dismissible"><p>' + r.data.message + '</p></div>').prependTo('.wrap');
+                                $('<div class="notice notice-success is-dismissible"><p>' + r.data.message + '</p></div>').insertAfter('.wp-header-end');
                             } else {
                                 $btn.prop('disabled', false).text(bedrock_ajax.add_text);
                                 alert('Error: ' + r.data.message);
--- a/web/app/mu-plugins/bedrock-composer-store.php
+++ b/web/app/mu-plugins/bedrock-composer-store.php
@@ -61,51 +61,73 @@ class BedrockComposerStore
         </style>
         <script>
             jQuery(function($) {
-                const plugins = [];
-                $('.plugin-card').each(function() {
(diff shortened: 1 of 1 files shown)


retlehs:
Wouldn't your lock file would become stale immediately after using this? This isn't something we'd consider adding to Bedrock, but I'd encourage you to share it on https://discourse.roots.io/ so others from the community could see it

FYI admin-ajax is extremely outdated, use the REST API which is made for more modern WP development

CallMeLeon167:
retlehs thanks for your input.

I think that would be a good extension for some users. I will create a package and share it with the community.

PR #781, opened 2025-09-16
fix: $webroot_dir phpdoc type

LucasDemea:
This fixes the following phpstan error regarding the type of $webroot_dir declared in the phpdoc :
```
PHPDoc tag @var with type string is not subtype of native type  
         non-falsy-string.                                               
         🪪  varTag.nativeType        
```


retlehs (approved):
Thank you!

